Issue #938, opened 2019-08-01
`_ImageNameParser.parse_remote()`: does not raise for unknown registry_url with port


ayushkovskiy:
During my work on PR https://github.com/neuromation/platform-client-python/pull/939, I found the following issue:

I think this branch is incorrect:
https://github.com/neuromation/platform-client-python/blob/a278ba4ec85d2cde25e7ae1a74d10641e0af2155/neuromation/api/parsing_utils.py#L78-L79

Instead, this method should look like:
```
    def parse_remote(self, value: str) -> RemoteImage:
        if not self.is_in_neuro_registry(value):
            raise ValueError("scheme 'image://' is required") # meaning "Image is expected to be in neuro registry"
        return self.parse_as_neuro_image(value)
```
because there's no reason to return then `RemoteImage` without fields `owner` and `registry` even if we already know the default values.

Test that does not work without this fix: case with registry-url that has port defined, when scheme is not defined: `"localhost:9999/bob/library/ubuntu:v10.04"`:

```
    def test_parse_remote__registry_has_port__image_in_bad_repo(self) -> None:
        my_parser = _ImageNameParser(
            default_user="alice", registry_url=URL("http://localhost:5000")
        )
        image = "localhost:9999/bob/library/ubuntu:v10.04"
        with pytest.raises(ValueError, match="scheme 'image://' is required"):
            my_parser.parse_remote(image)
```

@asvetlov please correct me if I'm wrong

PR #939, opened 2019-08-01
Support remote images with registry `host:port`

ayushkovskiy:
The newest version of neuro-API `19.7.26` broke platform-monitoring: https://github.com/neuromation/platform-monitoring/pull/70

The problem was that if `registry_url` contains both host and port, still `_ImageNameParser._registry` will contain only `host` part. Thus, parser `_container_from_api`, used to parse the response for  `Jobs.run`, will try to parse an image like `localhost:5000/user/ubuntu:latest` as a local image not as remote (because it does not start with `localhost/` but starts with `localhost:5000/`).

Note: one test is still xfailed due to https://github.com/neuromation/platform-client-python/issues/938
 (it fails with a wrong message)

### DIFF
--- a/neuromation/api/parsing_utils.py
+++ b/neuromation/api/parsing_utils.py
@@ -133,7 +133,14 @@ class _ImageNameParser:
         if not self.is_in_neuro_registry(image):
             raise ValueError("scheme 'image://' is required for remote images")
 
+        allow_port = False
         url = URL(image)
+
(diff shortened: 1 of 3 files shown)


ayushkovskiy:
I will add the changelog message to the second PR that resolves this issue: https://github.com/neuromation/platform-client-python/pull/940

PR #940, opened 2019-08-01
`_ImageNameParser.parse_remote()`: do not raise for unknown registry_url with port

ayushkovskiy:
Closes https://github.com/neuromation/platform-client-python/issues/938

### DIFF
--- a/neuromation/api/parsing_utils.py
+++ b/neuromation/api/parsing_utils.py
@@ -165,8 +165,12 @@ class _ImageNameParser:
     def _split_image_name(
         self, image: str, default_tag: Optional[str] = None
     ) -> Tuple[str, Optional[str]]:
+        if image.endswith(":") or image.startswith(":"):
+            # case `ubuntu:`, `:latest`
+            raise ValueError("empty name or empty tag")
         colon_count = image.count(":")
         if colon_count == 0:
+            # case `ubuntu`
             name, tag = image, default_tag
         elif colon_count == 1:
             # case `ubuntu:latest`
@@ -174,21 +178,16 @@ class _ImageNameParser:
             if "/" in tag:
                 # case `localhost:5000/ubuntu`
                 name, tag = image, default_tag
-            elif not tag:
-                raise ValueError("empty tag is not allowed")
         elif colon_count == 2:
             # case `localhost:9000/owner/ubuntu:latest`
             if "/" not in image:
-                # case `ubuntu:v11:latest`
+                # case `localhost:9000:latest`
                 raise ValueError("too many tags")
-            *image_arr, tag = image.split(":")
-            name = ":".join(image_arr)
+            name, tag = image.rsplit(":", 1)
         else:
             raise ValueError("too many tags")
-        if not tag:
-            tag = default_tag
-        elif "/" in tag:
-            raise ValueError("invalid tag format")
+        if tag and any(char in tag for char in "/:"):
+            raise ValueError("invalid tag")
         return name, tag
 
     def _check_allowed_uri_elements(self, url: URL) -> None:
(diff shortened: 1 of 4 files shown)


ayushkovskiy:
> What about the case localhost:9000/owner/ubuntu?

Thank you @serhiy-storchaka , this case is now also supported!

neuromation/api/parsing_utils.py
```diff
@@ -158,14 +165,25 @@ def _split_image_name(
...
+                name, tag = image, default_tag
+            elif not tag:
                 raise ValueError("empty tag is not allowed")
+        elif colon_count == 2:
+            # case `localhost:9000/owner/ubuntu:latest`
+            if "/" not in image:
+                # case `ubuntu:v11:latest`
+                raise ValueError("too many tags")
+            *image_arr, tag = image.split(":")
+            name = ":".join(image_arr)
```
serhiy-storchaka: It could be written shorter as `name, tag = image.rsplit(':', 1)`.

But what about "localhost/owner/ubuntu:v11:latest"? It is passed in contrary to "ubuntu:v11:latest".

Can a tag contain a slash? It is forbidden in some cases ("ubuntu:v18/04"), but is allowed in others ("localhost:9000/owner/ubuntu:v18/04").

Also the case of an empty tag is not handled here ("localhost:9000/owner/ubuntu:").

ayushkovskiy: > It could be written shorter as name, tag = image.rsplit(':', 1).

ok thanks!

> Can a tag contain a slash? It is forbidden in some cases ("ubuntu:v18/04"), but is allowed in others ("localhost:9000/owner/ubuntu:v18/04").

Indeed, docker forbids both cases:
```
$ docker pull ubuntu:10/04
Using default tag: latest
Error response from daemon: Get https://ubuntu:10/v2/: dial tcp: lookup ubuntu: no such host
$ docker pull localhost:5000/ubuntu:10/04
invalid reference format
```
added a check.

> Also the case of an empty tag is not handled here ("localhost:9000/owner/ubuntu:").

supported, thanks

ayushkovskiy:
By the way, similarly to `platform-monitoring`, we could use existing image format parser: `docker_image.reference.Reference.parse()`:
https://github.com/neuromation/platform-monitoring/blob/f4646efc5446020984b3ae045935757d73522837/platform_monitoring/docker_client.py#L61-L68

neuromation/api/parsing_utils.py
```diff
@@ -158,14 +165,29 @@ def _split_image_name(
...
         else:
             raise ValueError("too many tags")
-        return image, tag
+        if not tag:
+            tag = default_tag
```
serhiy-storchaka: "localhost/owner/ubuntu:" raises an error, but "localhost:9000/owner/ubuntu:" uses the default tag.

ayushkovskiy: considered, thanks

ayushkovskiy:
let's tuning docker image parser: there still are many cases not supported (for example, uppercase letters in image name) -- there's no need to implement a full parser, if something goes wrong, docker-server itself will reject the image.

neuromation/api/parsing_utils.py
```diff
@@ -163,30 +163,29 @@ def _parse_as_neuro_image(
...
-            *image_arr, tag = image.split(":")
-            name = ":".join(image_arr)
+            name, tag = image.rsplit(":", 1)
         else:
             raise ValueError("too many tags")
-        if not tag:
-            tag = default_tag
-        elif "/" in tag:
-            raise ValueError("invalid tag format")
+        if tag and any(char in tag for char in "/:"):
```
serhiy-storchaka: Would not
```python
if ":" in tag or "/" in tag:
```
be simple?

serhiy-storchaka: Can tag contain ":" at all?

ayushkovskiy: > Would not `if ":" in tag or "/" in tag:` be simple?
ok simplified

> Can tag contain ":" at all?
no, it can not

PR #946, opened 2019-08-05
Client-side support for streaming neuro-save

ayushkovskiy:


neuromation/api/jobs.py
```diff
@@ -162,14 +171,43 @@ def __init__(self, core: _Core, config: _Config) -> None:
             # an error is raised for status >= 400
             return None  # 201 status code
 
-    async def save(self, id: str, image: RemoteImage) -> None:
+    async def save(
```
asvetlov: Looking on dependency graph I'm inclining to a thought that this method actually belongs to images, not jobs.

neuromation/api/abc.py
```diff
@@ -99,10 +99,16 @@ class ImageProgressPush:
...
+class JobProgressCommit:
+    job: str
+    dst: RemoteImage
+
+
 @dataclass(frozen=True)
 class ImageProgressStep:
     message: str
-    layer_id: str
+    layer_id: str = ""
```
asvetlov: Why the attr becomes optional?
What is the use case for progress without layer?

neuromation/api/jobs.py
```diff
@@ -162,14 +171,43 @@ def __init__(self, core: _Core, config: _Config) -> None:
...
+        is_committing = True
+        is_pushing = False
+
         payload = {"container": {"image": _as_repo_str(image)}}
         url = self._config.cluster_config.monitoring_url / f"{id}/save"
-        async with self._core.request("POST", url, json=payload):
-            # an error is raised for status >= 400
-            return None  # 201 status code
+        async with self._core.request("POST", url, json=payload) as resp:
+            async for chunk in resp.content:
```
asvetlov: Please replace this loop with two sequential ones: first for committing, next for pushing.

neuromation/api/jobs.py
```diff
@@ -162,14 +171,43 @@ def __init__(self, core: _Core, config: _Config) -> None:
...
+                    if not is_pushing:
+                        # NOTE: here we don't expose our registry name in `src`
+                        src = LocalImage(f"{image.owner}/{image.name}", image.tag)
+                        progress.push(ImageProgressPush(src, dst=image))
+                        is_pushing = True
+                    step = _try_parse_image_progress_step(obj, image.tag)
+                    if step:
+                        progress.step(step)
+
+        return image
```
asvetlov: You don't modify `image` in the method, what is the reason for returning argument as is?
Return nothing is enough.

neuromation/cli/job.py
```diff
@@ -395,7 +396,7 @@ def job() -> None:
 async def _print_logs(root: Root, job: str) -> None:
     async for chunk in root.client.jobs.monitor(job):
         if not chunk:
-            break
+            continue
```
asvetlov: Is it related? Why?

neuromation/api/abc.py
ayushkovskiy: to use it for docker-commit messages as well: 
- `{"status": f"Committing container {cont_id} as image {name}:{tag}"}`
- `{"status": "Committed"}`

neuromation/api/jobs.py
ayushkovskiy: as far as I understood, the CLI command should be `neuro job save`, not `neuro image save` :confused:

neuromation/api/jobs.py
ayushkovskiy: OK thanks

neuromation/api/jobs.py
ayushkovskiy: ok

neuromation/cli/job.py
ayushkovskiy: oops.. sure, at least not in this PR

neuromation/api/abc.py
asvetlov: Please don't, you are mixing two independent message types.

neuromation/api/jobs.py
ayushkovskiy: How should it look like? if `self._core.request("POST", url, json=payload)` returns `asynccontextmanager`, which streams the following messages:
```
{"status": "Committing container DIGEST as image alpine:latest"},
        {"status": "Committed"},
        {"status": "The push refers to repository [localhost:5000/alpine]"},
        {"status": "Preparing", "progressDetail": {}, "id": "a31dbd3063d7"},
        {
            "status": "Pushing",
            "progressDetail": {"current": 3584},
            "progress": " 3.584kB",
            "id": "0acd017a4b67",
        },
        {"status": "Pushed", "progressDetail": {}, "id": "0acd017a4b67"},
        {"status": "job-id: digest: sha256:DIGEST size: 1359"},
        {
            "progressDetail": {},
            "aux": {"Tag": "job-id", "Digest": "sha256:DIGEST", "Size": 1359},
        },
```

neuromation/api/jobs.py
asvetlov: This is a good question.
In docker `save` belongs to `docker image` group, not `docker container`.
Maybe we should apply the same change to our CLI.

Anyway, for now please keep CLI command as is but move API.

neuromation/api/jobs.py
asvetlov: So you effectively have something like:
```
line1 = await resp.content.readline()  # Committing container
line2 = await resp.content.readline()  # Committed
line3 = await resp.content.readline()  # start pushing
async for line in resp.content:
    # preparing, pushing, pushed etc
```

neuromation/api/jobs.py
ayushkovskiy: ok

neuromation/api/jobs.py
ayushkovskiy: achso, okay thanks!

neuromation/api/abc.py
ayushkovskiy: ok

neuromation/api/abc.py
```diff
@@ -117,3 +128,13 @@ def push(self, data: ImageProgressPush) -> None:
     @abc.abstractmethod
     def step(self, data: ImageProgressStep) -> None:
         pass  # pragma: no cover
+
+
+class AbstractDockerContainerProgress(abc.ABC):
```
ayushkovskiy: `docker-commit` progress is now separated from `docker-push/pull`. Considering that there's only 2 messages to print all of format `{"status": "message..."}`, we can remove this abstraction.

neuromation/api/images.py
```diff
@@ -140,20 +144,65 @@ def _auth(self) -> Dict[str, str]:
...
+            image_progress = _DummyImageProgress()
+
+        payload = {"container": {"image": _as_repo_str(image)}}
+        url = self._config.cluster_config.monitoring_url / f"{id}/save"
+
+        async with self._core.request("POST", url, json=payload) as resp:
+
+            # first, we process docker-commit messages
+            container_progress.commit(ContainerProgressCommit(id, image))
+            async for chunk in resp.content:
```
ayushkovskiy: IMO cycle over docker-commit messages looks better than manual processing of 2 commit chunks:
```
# first, we expect exactly two docker-commit messages
container_progress.commit(ContainerProgressCommit(id, image))

commit_chunk_1 = self._load_chunk(await resp.content.readline())
step_1 = _try_parse_job_commit_progress_step(commit_chunk_1)
container_progress.step(step_1)

commit_chunk_2 = self._load_chunk(await resp.content.readline())
step_2 = _try_parse_job_commit_progress_step(commit_chunk_2)
container_progress.step(step_2)
if step_2.message != "Committed":
    error_details = {
        "message": f"Failed to commit: last chunk status: `{step_2.message}`"
    }
    raise DockerError(900, error_details)
...
```

neuromation/api/images.py
```diff
@@ -140,20 +144,65 @@ def _auth(self) -> Dict[str, str]:
...
+                obj = self._load_chunk(chunk)
+                commit_step = _try_parse_job_commit_progress_step(obj)
```
ayushkovskiy: AFAIK, docker-commit does not support streaming (even though it can take a while, for larger images). Thus, there always will be only 2 messages, "about to commit" and "committed", or an error message. So let's not print these messages at all, and just check if there's no errors?

ayushkovskiy: the code will be simplified significantly

neuromation/api/jobs.py
```diff
@@ -162,15 +157,6 @@ def __init__(self, core: _Core, config: _Config) -> None:
             # an error is raised for status >= 400
             return None  # 201 status code
 
-    async def save(self, id: str, image: RemoteImage) -> None:
```
ayushkovskiy: moved to api/images.py

neuromation/api/abc.py
ayushkovskiy: maybe, we can change this class to have two methods: 
- `start_commit()`, and
- `finish_commit()`

neuromation/api/images.py
asvetlov: Agree, the commit progress is pointless if docker doesn't provide *progress* information for it.

neuromation/api/abc.py
ayushkovskiy: no the case anymore

neuromation/api/jobs.py
ayushkovskiy: I moved both `save()` back to `jobs` because when it belongs to `images`, this functionality does not work on windows as `Images.__init__()` requires docker available on the local machine, which is not necessary for `save()`.

Let's keep it less clean for now, and later refactor it out to a separate class (a separate member of `Client`).

asvetlov: Separate member of `Client` sounds even better.
What name do you suggest?
What prevents you from making this refactoring now?

ayushkovskiy: > What name do you suggest

maybe something like `ImagesLocal` (for current `Images`) and `ImagesRemote`? meaning if the user's machine works with local docker or remote docker.

> What prevents you from making this refactoring now?

It will require many non-functional changes; let's keep this PR as-is and do refactoring it in a separate PR?

neuromation/api/abc.py
```diff
@@ -99,12 +101,35 @@ class ImageProgressPush:
...
+
+
+@dataclass(frozen=True)
+class ImageCommitDetails:
+    container: str
+    target_image: RemoteImage
+
+
+@dataclass(frozen=True)
+class ImageCommitStep:
```
asvetlov: Looks like you have always two `ImageCommitStep` messages: one with `status==CommitStarted`, second with `status=CommitFinished`.

Why not replace it with two distinct message type:
```
@dataclass
class ImageCommitStarted:
    pass  # really no context data? I expect job id at least

@class
class ImageCommitFinished:
    container: str
    target_image: RemoteImage
    # job id again, maybe?
```

ayushkovskiy: Honestly, I don't see the difference. Re-worked.

neuromation/api/abc.py
```diff
@@ -99,12 +99,30 @@ class ImageProgressPush:
...
 @dataclass(frozen=True)
 class ImageProgressStep:
     message: str
     layer_id: str
 
 
+@dataclass(frozen=True)
+class ImageCommitStarted:
+    job_id: str
+    container: str
```
asvetlov: What `container` string is?
In jobs API we have complex `Container` dataclass. I have an understanding of what dataclass fields mean. We'll have documentation for it soon.
As a user, I have no idea what to do with container-as-string except printing it may be (but what is the string format)?

ayushkovskiy: it's 64-long SHA256 hash of docker container

ayushkovskiy: mostly technical information, but since we reveal operations "commit" and "push", I decided to print it as well.

asvetlov: Hash of docker container or image?
If this is a local docker container hash -- how can I use this info as a client?
Client has no access to docker daemon deployed on the node.

ayushkovskiy: hash of docker container. 

> Client has no access to docker daemon deployed on the node.

 true. I will remove `container: str`.

ayushkovskiy: In this case, we also should not print `Pushing image artemyushkovskiy/test:latest => image://artemyushkovskiy/test:latest` because the source image does not  make sense for the user

ayushkovskiy: I removed messages about "container", but I will not fix docker-push messages `"Pushing image artemyushkovskiy/test:latest => image://artemyushkovskiy/test:latest"` in this PR as it is a very minor issue that requires lots of changes.

PR #950, opened 2019-08-07
Drop JobDescriptor.http_url_named

asvetlov:
Support for both `JobDescriptor.http_url_named` and `JobDescriptor.http_url` in API is pointless: the code always uses *named* version if present, unnamed otherwise.

`http_url_named` should be dropped from public API at least, if server returns `http_url_named` in result JSON it should be converted to `http_url` implicitly.

There is no sense for documenting both properties or using them by user code.

### DIFF
--- a/neuromation/cli/formatters/jobs.py
+++ b/neuromation/cli/formatters/jobs.py
@@ -48,7 +48,7 @@ class JobFormatter:
             job_alias = job.name
         else:
             job_alias = job.id
-        http_url = job.http_url_named or job.http_url
+        http_url = job.http_url
         if http_url:
             out.append(style("Http URL", bold=True) + f": {http_url}")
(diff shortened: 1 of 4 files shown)


neuromation/cli/formatters/jobs.py
```diff
@@ -90,7 +90,7 @@ def __call__(self, job_status: JobDescription) -> str:
         result += f"Preemptible: {job_status.is_preemptible}\n"
         if job_status.internal_hostname:
             result += f"Internal Hostname: {job_status.internal_hostname}\n"
-        http_url = job_status.http_url_named or job_status.http_url
+        http_url = job_status.http_url
```
serhiy-storchaka: I think it can be inlined, as `job_status.internal_hostname` or other attributes.

asvetlov: done

PR #951, opened 2019-08-07
Drop obsolete and unused JobDescriptor.ssh_server

asvetlov:
The attribute used to support the old implementation of `neuro exec` but this code is dropped by the client now.
There is no sense for keeping this attr now.

### DIFF
--- a/neuromation/api/jobs.py
+++ b/neuromation/api/jobs.py
@@ -94,7 +94,6 @@ class JobDescription:
     description: Optional[str] = None
     http_url: URL = URL()
     http_url_named: URL = URL()
-    ssh_server: URL = URL()
     internal_hostname: Optional[str] = None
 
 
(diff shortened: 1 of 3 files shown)


serhiy-storchaka:
Actually `ssh_auth_server` is deprecated in favor of `ssh_server`. See neuromation/platform-api#575.

asvetlov:
Funny, `ssh_server` is not used by neuromation client at all but `ssh_auth_server` is used by exec: https://github.com/neuromation/platform-client-python/blob/master/neuromation/api/jobs.py#L240 and port-forward: https://github.com/neuromation/platform-client-python/blob/master/neuromation/api/jobs.py#L276

Do you suggest replacing these lines with non-deprecated `ssh_server`?

asvetlov:
#952 provides the opposite solution

Issue #942, opened 2019-08-02
Invert interactivity default for `neuro exec`


mariyadavydova:
Make it work interactively without -t option, so that `neuro exec <job> bash` is interactive by default.

PR #957, opened 2019-08-14
Invert interactivity default for `neuro exec`

ayushkovskiy:
1. Closes https://github.com/neuromation/platform-client-python/issues/942
2. Fixes e2e tests cleanup

### DIFF
--- a/neuromation/cli/job.py
+++ b/neuromation/cli/job.py
@@ -279,11 +279,11 @@ async def submit(
 @click.argument("job")
 @click.argument("cmd", nargs=-1, type=click.UNPROCESSED, required=True)
 @click.option(
-    "-T",
-    "--no-tty",
-    default=False,
+    "-t/-T",
+    "--tty/--no-tty",
+    default=True,
     is_flag=True,
-    help="Do not allocate a virtual tty. Should be used for non-interactive commands.",
+    help="Allocate virtual tty. Useful for interactive jobs.",
 )
 @click.option(
     "--no-key-check",
@@ -301,7 +301,7 @@ async def submit(
 async def exec(
     root: Root,
     job: str,
-    no_tty: bool,
+    tty: bool,
     no_key_check: bool,
     cmd: Sequence[str],
     timeout: float,
@@ -315,15 +315,14 @@ async def exec(
     neuro exec my-job /bin/bash
 
     # Executes a single command in the container and returns the control:
-    neuro exec --no-tty my-job ls > ls_output.txt && echo "OK"
-    neuro exec -T my-job /bin/not-an-executable || echo "failed"
+    neuro exec --no-tty my-job ls -l
     """
     cmd = shlex.split(" ".join(cmd))
     id = await resolve_job(root.client, job)
     retcode = await root.client.jobs.exec(
         id,
         cmd,
-        tty=not no_tty,
+        tty=tty,
         no_key_check=no_key_check,
         timeout=timeout if timeout else None,
     )
--- a/neuromation/cli/job.py
+++ b/neuromation/cli/job.py
@@ -277,12 +277,13 @@ async def submit(
 
 @command(context_settings=dict(allow_interspersed_args=False))
 @click.argument("job")
-@click.argument("cmd", nargs=-1, type=click.UNPROCESSED)
+@click.argument("cmd", nargs=-1, type=click.UNPROCESSED, required=True)
 @click.option(
-    "-t",
-    "--tty",
+    "-T",
+    "--no-tty",
+    default=False,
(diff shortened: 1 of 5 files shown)


tests/e2e/conftest.py
```diff
@@ -436,7 +442,7 @@ def run_cli(self, arguments: List[str], *, verbosity: int = 0) -> SysCap:
                 sleep(delay)
                 delay *= 2
                 continue
-            else:
+            elif wait_for_exit_code:
```
ayushkovskiy: sometimes, when we execute an interactive command, we don't want to wait until it returns (it won't)

tests/e2e/conftest.py
```diff
@@ -446,7 +452,7 @@ def run_cli(self, arguments: List[str], *, verbosity: int = 0) -> SysCap:
             out = proc.stdout
             err = proc.stderr
             if any(
-                " ".join(arguments).startswith(start)
+                start in " ".join(arguments)
```
ayushkovskiy: This fixes the cleanup broken because of global options passed (e.g., `neuro -q job submit`)

tests/e2e/test_e2e_jobs.py
```diff
@@ -591,6 +591,22 @@ def test_e2e_ssh_exec_tty(helper: Helper) -> None:
     assert captured.out == ""
 
 
+@pytest.mark.e2e
+def test_e2e_ssh_exec_tty_is_by_default(helper: Helper) -> None:
```
ayushkovskiy: since exec tests are quite flaky, maybe we might want to consider removing (or at least muting) the other test: `test_e2e_ssh_exec_tty`

neuromation/cli/job.py
```diff
@@ -307,7 +307,22 @@ def job() -> None:
...
+
+    # Provides a shell to the container via `/bin/bash`:
+    neuro exec my-job
+    neuro exec --tty my-job /bin/bash
+
+    # Executes a single command in the container and returns the control:
+    neuro exec my-job ls > ls_output.txt && echo "OK"
+    neuro exec my-job /bin/not-an-executable || echo "failed"
     """
+    # TODO (artem) Provide a backup option if the container lacks `/bin/bash`
```
ayushkovskiy: something like:
try: run `/bin/bash`, except executable-not-found-error: run `/bin/sh` ...

But definitely later, not in this PR. Probably this concern is pointless since (1) most of data scientists' containers are based on ubuntu or are reach and have `/bin/bash` anyhow, and (2) we explicitly stated that by default we run `/bin/bash` in the help message, so the user is aware.

ayushkovskiy:
Please note that e2e tests flake a lot. 
Successful builds for the [last commit](https://github.com/neuromation/platform-client-python/pull/957/commits/91d5562137896fff7d4829099fa5b7acc224322f):
- `build_py36` https://circleci.com/gh/neuromation/platform-client-python/5710
- `windows, py36+py37` https://ci.appveyor.com/project/astaff/platform-api-clients/builds/26699398

ayushkovskiy:
@asvetlov Thanks for your comments, all fixed.

neuromation/cli/job.py
```diff
@@ -279,10 +279,11 @@ def job() -> None:
 @click.argument("job")
 @click.argument("cmd", nargs=-1, type=click.UNPROCESSED, required=True)
 @click.option(
-    "-t",
-    "--tty",
+    "-T",
```
asvetlov: Please use `"-t/-T", "--tty/--no-tty"` form.
I'd like to keep `--tty` flag at least for backward compatibility but make it default.

neuromation/cli/job.py
```diff
@@ -300,20 +301,29 @@ def job() -> None:
...
     """
     Execute command in a running job.
+
+    Examples:
+
+    # Provides a shell to the container:
+    neuro exec my-job /bin/bash
+
+    # Executes a single command in the container and returns the control:
+    neuro exec --no-tty my-job ls > ls_output.txt && echo "OK"
```
asvetlov: Please make examples simpler, avoid shell redirection and command chaining if possible.

neuromation/cli/job.py
```diff
@@ -300,20 +301,29 @@ def job() -> None:
...
+    neuro exec -T my-job /bin/not-an-executable || echo "failed"
```
asvetlov: Providing an example for executing non-existing command looks not very informative.

neuromation/cli/job.py
ayushkovskiy: ok sure

neuromation/cli/job.py

Issue #943, opened 2019-08-02
Add an info message in `neuro run`/`neuro submit`


mariyadavydova:
The message should appear in case of no `--detach` option is used. It should tell something like this:

```
This terminal is attached to the remote job, so you receive the job output.
Use 'Ctrl-C' to detach from this job (this won't terminate the job),
or restart the job with `--detach` option. 
```

PR #962, opened 2019-08-19
Add an info message in `neuro run`/`neuro submit`

ayushkovskiy:
Closes https://github.com/neuromation/platform-client-python/issues/943

### DIFF
--- a/neuromation/cli/job.py
+++ b/neuromation/cli/job.py
@@ -3,6 +3,7 @@ import logging
 import os
 import shlex
 import sys
+import textwrap
 import uuid
 import webbrowser
 from typing import Dict, List, Optional, Sequence, Set, Tuple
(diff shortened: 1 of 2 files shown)


neuromation/cli/job.py
```diff
@@ -853,6 +853,12 @@ def format_fail(job: str, reason: Exception) -> str:
         await browse_job(root, job)
 
     if not detach:
+        click.echo(
+            "This terminal is attached to the remote job, so you receive "
+            "the job output.\nUse 'Ctrl-C' to detach from this job (it "
```
serhiy-storchaka: Could it be reformatted so that lines will break after `\n`?

ayushkovskiy: done, thanks

asvetlov:
@ayushkovskiy what do you think about my suggestion for gray color (`dim=True`)?

ayushkovskiy:
> what do you think

@asvetlov just pushed :wink:

Issue #944, opened 2019-08-02
Show certifi package error not earlier than two weeks after the latest version release


mariyadavydova:
Two weeks gap is approximate time for this package to reach conda and other package managers after pip release.
This gap will decrease the time when a user sees this error message but can't get rid of it as certifi is installed by the package manager, not pip.

ayushkovskiy:
Is this a common practice, to mute errors with empirically defined delay?

@asvetlov, @serhiy-storchaka please share your thoughts

asvetlov:
Sorry for the delay.
I never seem such logic before but suggestion from @mariyadavydova makes sense.
There is a gap between publishing a new version on PyPI and mirroring it by conda.
Just after PyPI release there is no update on conda, it appears somewhen a week later.


`https://pypi.org/pypi/certifi/json` provides `"upload_time": "2016-02-28T14:32:48"` for every release.
6-month-old `certifi` is a security whole. 2 weeks old is not a problem.
We can just relax a warning a little.

ayushkovskiy:
perfect. My implementation also depends on `"upload_time"`, the PR will be soon.

My concern is that this value almost literally repeats `certifi_pypi_version` since certifi uses calver, thus we could simplify logic significantly and avoid reading and storing in `.nmrc` an additional parameter `certifi_pypi_upload_time`.

asvetlov:
I feel that `upload_time` is better from a code reading perspective, but I don't care too much

PR #970, opened 2019-08-21
Show certifi package error not earlier than two weeks after the latest version release

ayushkovskiy:
Closes https://github.com/neuromation/platform-client-python/issues/944

### DIFF
--- a/neuromation/api/config.py
+++ b/neuromation/api/config.py
@@ -55,7 +55,7 @@ class _PyPIVersion:
         )
 
     def to_config(self) -> Dict[str, Any]:
-        res = {
+        ret = {
             "pypi_version": str(self.pypi_version),
             "check_timestamp": int(self.check_timestamp),
             "certifi_pypi_version": str(self.certifi_pypi_version),
@@ -63,9 +63,9 @@ class _PyPIVersion:
         }
         if self.certifi_pypi_upload_date != date.min:
             value = self._serialize_date(self.certifi_pypi_upload_date)
-            res["certifi_pypi_upload_date"] = value
+            ret["certifi_pypi_upload_date"] = value
 
-        return res
+        return ret
 
     @classmethod
     def _deserialize_date(cls, value: str) -> date:
--- a/neuromation/api/config.py
+++ b/neuromation/api/config.py
@@ -1,6 +1,8 @@
 from dataclasses import dataclass
+from datetime import date
 from typing import Any, Dict
 
+import dateutil.parser
 import pkg_resources
 from yarl import URL
 
@@ -15,10 +17,11 @@ class _PyPIVersion:
     check_timestamp: int
     certifi_pypi_version: Any
     certifi_check_timestamp: int
+    certifi_pypi_upload_date: date = date.min
 
     @classmethod
     def create_uninitialized(cls) -> "_PyPIVersion":
-        return cls(cls.NO_VERSION, 0, cls.NO_VERSION, 0)
+        return cls(cls.NO_VERSION, 0, cls.NO_VERSION, 0, date.min)
 
     @classmethod
     def from_config(cls, data: Dict[str, Any]) -> "_PyPIVersion":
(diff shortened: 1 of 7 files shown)


neuromation/api/config.py
```diff
@@ -33,25 +36,46 @@ def from_config(cls, data: Dict[str, Any]) -> "_PyPIVersion":
...
-        return {
+        res = {
             "pypi_version": str(self.pypi_version),
             "check_timestamp": int(self.check_timestamp),
             "certifi_pypi_version": str(self.certifi_pypi_version),
             "certifi_check_timestamp": int(self.certifi_check_timestamp),
         }
+        if self.certifi_pypi_upload_date != date.min:
+            value = self._serialize_date(self.certifi_pypi_upload_date)
+            res["certifi_pypi_upload_date"] = value
```
ayushkovskiy: note: since certifi uses [calver](https://calver.org), there will be a duplicating information in `.nmrc`:
```
pypi:
  certifi_pypi_upload_date: '2019-06-16'
  certifi_pypi_version: 2019.6.16
  ...
```

tests/api/test_config.py
```diff
@@ -126,3 +129,69 @@ def test_check_initialized_bad_cluster_config(self) -> None:
         )
         with pytest.raises(ValueError, match="Missing server configuration"):
             config.check_initialized()
+
+
+class TestPyPIVersion:
```
ayushkovskiy: I couldn't find any unit test on `_PyPIVersion` so created a new one

tests/cli/test_utils.py
```diff
@@ -339,3 +342,37 @@ def test_local_remote_port_param_type_invalid(arg: str) -> None:
     param = LocalRemotePortParamType()
     with pytest.raises(click.BadParameter, match=".* is not a valid port combination"):
         param.convert(arg, None, None)
+
+
+class TestDelayedReleaseWarning:
+    date_to_patch = f"{_need_to_warn_after_delay.__module__}.date"
```
ayushkovskiy: better to use the package [freezegun](https://pypi.org/project/pytest-freezegun/), but it requires an additional test dependency:
```
@pytest.mark.freeze_time('2017-05-21')
def test_current_date(current_date):
    pass
```

tests/cli/test_version_check.py
```diff
@@ -242,8 +263,11 @@ def pypi_server(fake_pypi: Tuple[FakePyPI, Dict[str, int]]) -> FakePyPI:
     async with VersionChecker(
         _PyPIVersion.create_uninitialized(), connector=connector
     ) as checker:
-        version = await checker._fetch_pypi("neuromation")
+        payload = await checker._fetch_pypi("neuromation")
+        version = checker._parse_max_version(payload)
         assert version == pkg_resources.parse_version("0.0.0")
+        upload_time = checker._parse_version_upload_time(payload, version)
+        assert upload_time == date.min
```
ayushkovskiy: In this PR, we miss a test that checks that when `await checker.run()` is called, then indeed all versions get updated. Will be in a consequent PR (already implemented).

tests/api/test_config.py
asvetlov: There is `cli/test_version_check.py` but it is a little different.

tests/cli/test_utils.py
asvetlov: freezegun has own limitations: it doesn't shift the process time but only some Python APIs.
Thus tested code becomes desynchronized easily and quickly.

If you really want to check `datetime` module operations correctness you can pass `now` into `_need_to_warn_after_delay()` helper. 
Quite pointless work though.

ayushkovskiy: if it's pointless, should we remove the whole bunch of tests? indeed the logic is very simple there

asvetlov: Please remove.
If we change these simple lines patching mocked tests can be tedious.
A chance to make an error in `_need_to_warn_after_delay()` is neglible.

tests/api/test_config.py
ayushkovskiy: moved there.

ayushkovskiy:
comments by @asvetlov considered, thank you.

asvetlov:
LGTM on green

Issue #971, opened 2019-08-21
Image parsing fallback


shagren:
If you having invalid images like `image://shagren/:latest` which are incorrect but persist in you job history then latest `neuromation cli` will fail on `ps -a` command with error like:
```
  Invalid remote image 'dev.neu.ro/shagren/:latest': empty name or empty tag
```

Let's add some UnknownImage/UnparsableImage class for handling this case for `ps` / `status` commands.

asvetlov:
Thanks for the report.
We should construct something like `Invalid Image Name` [here] (https://github.com/neuromation/platform-client-python/blob/master/neuromation/api/jobs.py#L363-L372) if `parse_image()` call raises exception.

dalazx:
addressed by the referenced PR.

PR #972, opened 2019-08-21
Allow to parse legacy invalid images for ps/status commands

shagren:
refs #971

### DIFF
--- a/neuromation/api/jobs.py
+++ b/neuromation/api/jobs.py
@@ -5,17 +5,7 @@ import shlex
 from contextlib import suppress
 from dataclasses import dataclass, field
 from datetime import datetime
-from typing import (
-    Any,
-    AsyncIterator,
-    Dict,
-    List,
-    Mapping,
-    Optional,
-    Sequence,
-    Set,
-    Union,
-)
+from typing import Any, AsyncIterator, Dict, List, Mapping, Optional, Sequence, Set
 
 import async_timeout
 import attr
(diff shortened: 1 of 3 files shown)


neuromation/api/jobs.py
```diff
@@ -361,8 +361,13 @@ def _http_port_from_api(data: Dict[str, Any]) -> HTTPPort:
 
 
 def _container_from_api(data: Dict[str, Any], parser: _ImageNameParser) -> Container:
+    try:
+        image = parser.parse_remote(data["image"])
+    except ValueError:
+        image = RemoteImage(name=f'Invalid Image Name {data["image"]}')
```
asvetlov: I'm curious how this super-long-line looks in tables view provided by `TabularJobsFormatter`?

shagren: It is why I propose to add new Class, then we can use different formatting for `ps` and `status`.
I.e. `InvalidImage` for `ps` and `Invalid Image Name [Name here]` for `ps`

asvetlov: Do you really want to print this invalid image name?
Now you cannot create such a thing, the only way to get invalid image is retrieving historical old enough job. Missing exact invalid image name doesn't make much harm.
I don't insist though. If you prefer your version -- please go ahead and merge.

The very minor and rare thing anyway.

shagren: Agree, will change

Issue #978, opened 2019-08-22
port-forward isn't working


Rauf-Kurbanov:
```
➜  neuro port-forward dev 1489:22
Your root certificates are out of date.
You are using certifi 2019.3.9, however 2019.6.16 is available.
Please consider upgrading certifi package, e.g.
    pip install --upgrade certifi
or
    conda update certifi
Port localhost:1489 will be forwarded to port 22 of job-d74c8c92-dddf-4db6-bb28-62d297bb48d6
Press ^C to stop forwarding
➜  ps aux | grep ssh
raufkurbanov      8750   0.0  0.0  4278260    796 s001  R+    6:59PM   0:00.00 grep --color=auto --exclude-dir=.bzr --exclude-dir=CVS --exclude-dir=.git --exclude-dir=.hg --exclude-dir=.svn ssh
raufkurbanov      7276   0.0  0.0  4289444   4324   ??  S     6:08PM   0:00.03 /usr/bin/ssh-agent -l
➜  neuro --version
Neuromation Platform Client 19.8.19
```

Rauf-Kurbanov:
works fine with `neuromation==19.8.1`

PR #979, opened 2019-08-22
Hotfix of port forwarding

ayushkovskiy:
Closes https://github.com/neuromation/platform-client-python/issues/978

### DIFF
--- a/neuromation/cli/job.py
+++ b/neuromation/cli/job.py
@@ -376,6 +376,11 @@ async def port_forward(
             )
 
         click.echo("Press ^C to stop forwarding")
+        try:
+            await asyncio.sleep(float('inf'))
+        except KeyboardInterrupt:
+            pass
+
 
 
 @command()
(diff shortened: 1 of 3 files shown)


neuromation/cli/job.py
```diff
@@ -376,6 +376,11 @@ def job() -> None:
             )
 
         click.echo("Press ^C to stop forwarding")
+        try:
+            await asyncio.sleep(float('inf'))
```
ayushkovskiy: @asvetlov please correct me to use a better infinity-waiting code

asvetlov:
Sorry, it is my fault.

neuromation/cli/job.py
asvetlov: Strictly speaking, it is not easy with current asyncio.
More or less robust way in the explicit SIGINT/SIGTERM handling.

Anyway, if you want to support Windows for Python 3.6 and 3.7 please replace infinite sleep with infinite loop of one-second periods of sleeping.

Issue #974, opened 2019-08-22
Implement mounting all available volumes


dalazx:
similarly to `--volume=HOME` we might want to introduce another option: `--volume=ALL`.

it would mount all the storage URIs available to the user under the known root directory.

1. Expose an env var denoting the root;
2. Request `neuro acl list` for the storage scheme via API;
3. Define the volumes list with the appropriate R/W mode;

Some question:
How would `ALL` play with `HOME`?
How would `ALL` play with `--pass-config`?

ayushkovskiy:
> How would ALL play with HOME?
k> How would ALL play with --pass-config?

for now, let's forbid both

ayushkovskiy:
Blocked by: https://github.com/neuromation/platform-client-python/issues/975

PR #977, opened 2019-08-22
Fix mounts for `--volume=HOME`

ayushkovskiy:
Closes https://github.com/neuromation/platform-client-python/issues/975

PR #980, opened 2019-08-22
Automount: neuro run --volume=ALL

ayushkovskiy:
closes https://github.com/neuromation/platform-client-python/issues/974


asvetlov:
Please merge #977, rebase against the master, and ask for review again.
Keeping in mind what is related to the PR and what not is a challenge. I would like to not resolve this pussle.

ayushkovskiy:
> Please merge #977, rebase against the master, and ask for review again.

Lately, when I was having multiple consequent PRs one depending on another, I always added the notification "WARNING: depends on #977" to the description of all non-ready-for-review PRs. Sorry for misunderstanding. From now on, I will not open a non-WIP PRs if they are not ready to be compared against master.

tests/cli/test_job.py
```diff
@@ -96,3 +97,33 @@ def test_calc_statuses__check_defaults__all_statuses_true(
     assert not std.out
     assert not std.err
     assert not caplog.text
+
+
+@pytest.mark.parametrize(
+    "env_var", ["NP_STORAGE_ROOT_MOUNTPOINT", "NP_STORAGE_ROOT_MOUNTPOINT=value"]
+)
+def test_build_env_reserved_env_var_conflict_passed_as_parameter(env_var: str) -> None:
```
ayushkovskiy: apart from these two tests, `build_env` is not covered

neuromation/cli/job.py
```diff
@@ -59,6 +59,9 @@
 
 log = logging.getLogger(__name__)
 
+ENV_VAR_STORAGE_ROOT_MOUNTPOINT = "NP_STORAGE_ROOT_MOUNTPOINT"
```
dalazx:
```suggestion
ENV_VAR_STORAGE_ROOT_MOUNTPOINT = "NP_STORAGE_ROOT"
```
or `NP_STORAGE_PATH` to make it shorter.

neuromation/cli/job.py
```diff
@@ -877,6 +873,55 @@ def format_fail(job: str, reason: Exception) -> str:
...
+                volumes.add(
+                    root.client.parse.volume(
+                        "storage://neuromation/public:/var/storage/neuromation:ro"
+                    )
+                )
+            if vol == "ALL":
+                available = await root.client.users.get_acl(
+                    root.username, scheme="storage"
+                )
+                root_mountpoint = "/var/storage"
```
dalazx: should we extract a constant and reuse throughout?
we could expose an env var to override the value (for the `HOME` as well).

neuromation/cli/job.py
```diff
@@ -877,6 +873,55 @@ def format_fail(job: str, reason: Exception) -> str:
     return job
 
 
+async def _build_volumes(
+    root: Root, input_volumes: Sequence[str], env_dict: Dict[str, str]
```
dalazx: are `input_volumes` unique?

neuromation/cli/job.py
asvetlov: Why do you need the env var at all?

neuromation/cli/job.py
```diff
@@ -877,6 +873,55 @@ def format_fail(job: str, reason: Exception) -> str:
...
+                root_mountpoint = "/var/storage"
+                env_dict[ENV_VAR_STORAGE_ROOT_MOUNTPOINT] = root_mountpoint
+                if not root.quiet:
+                    click.echo(
+                        "Storage root mountpoint will be available as the environment "
+                        f"variable: {ENV_VAR_STORAGE_ROOT_MOUNTPOINT}={root_mountpoint}"
+                    )
+                for perm in available:
+                    root_dir = (
+                        perm.uri.host if perm.uri.host != root.username else "home"
```
dalazx: not sure the `home` link is what we want really. I'd rather have a separate link (an additional mount) instead of renaming the owners home dir. also, we could easily introduce an environment var to address this (e.g. `$HOME`).

tests/cli/test_job.py
dalazx: should we properly cover it then?

neuromation/cli/job.py
dalazx: we need an env variable here to communicate to a user that the storage tree would be available under a certain directory and be inferred programmatically.

asvetlov: We use `/storage/home` constant for `HOME` processing, nobody complains.
Let's use the same strategy.
I prefer to avoid configuration options if they are not requested.

Also, the only already used/provided env var is `NEUROMATION_CONFIG` so far.
`NP_STORAGE_ROOT_MOUNTPOINT` doesn't look consistent.
I know we use `NP_` prefix on *server-side* but people see *client* API/CLI only

dalazx: it is not an env var for `neuro`. it is an env var for jobs.
so, we are not exposing another configuration option, but rather informing users for programmatic usage options.  (same as with `$HOME`).

as the matter of fact, `NP_` was originally used as the platform env var prefix for platform-specific env vars inside jobs. let's rename to whatever is considered consistent.

asvetlov: Thank you for correction.
I'm curious, is there a case when a mounting point is not `/var/storage`?
As a user, I prefer to use string constants over env vars if these vars never changes.

Don't get me wrong, I can live with this envvar. The question is: do we need to teach our clients about it?

dalazx: it seems in both cases we will teach users either to hardcode `/var/storage` or hardcode `NEUROMATION_STORAGE_ROOT`. the latter is advantageous to us as we may easily change the value.

asvetlov: Ok, let's add a variable that is ignored by most users :)
`NEUROMATION_STORAGE_ROOT` looks good to me.
In this PR, `HOME` is exposed as `${NEUROMATION_STORAGE_ROOT}/home`.
Should we introduce another env var or hardcoded `home` dirname is ok?

tests/cli/test_job.py
ayushkovskiy: sure. but not in this PR. Issue: https://github.com/neuromation/platform-client-python/issues/995

neuromation/cli/job.py

ayushkovskiy: Personally, I like the idea of exposing `NEUROMATION_STORAGE_ROOT=/var/storage` and re-using it for both `${NEUROMATION_STORAGE_ROOT}/home` and all other mounts.

neuromation/cli/job.py
ayushkovskiy: Thanks, for now, I moved this variable to the top of the file. In future, if needed, we always can expose it as an env var.

neuromation/cli/job.py
ayushkovskiy: no, it's direct user input.

neuromation/cli/job.py
ayushkovskiy: Please note that the method `upload_and_map_config` creates directory `/var/storage/nmrc`, so if a user has directory `storage://~/nmrc`, they will be in conflict. Let's maybe change `local_nmrc_folder` to `/var/storage/.nmrc`?

neuromation/cli/job.py
ayushkovskiy: In this PR, we made this decision to mount user home directory to `/var/storage/home` not to `/var/storage/username` in order to be compliant with `--volume=HOME` in order not to confuse users. I don't think we need to have both mounts.

ayushkovskiy: In addition, in a scheme `storage://username/path/to/dir`, we do not use `username` as a location of user's home directory, technically, in NFS it can be anywhere (for example, not in `/username` but `/home/username`), so I think it's good practice to mount user dir to `/var/storage/home`.

neuromation/cli/job.py
```diff
@@ -877,14 +875,64 @@ def format_fail(job: str, reason: Exception) -> str:
...
+
 async def upload_and_map_config(root: Root) -> Tuple[str, Volume]:
 
     # store the Neuro CLI config on the storage under some random path
     nmrc_path = URL(root.config_path.expanduser().resolve().as_uri())
     random_nmrc_filename = f"{uuid.uuid4()}-nmrc"
     storage_nmrc_folder = f"storage://{root.username}/nmrc/"
     storage_nmrc_path = URL(f"{storage_nmrc_folder}{random_nmrc_filename}")
-    local_nmrc_folder = "/var/storage/nmrc/"
+    # TODO (artem) if user has dir `storage:nmrc`, `--volume=ALL` will cause a conflict
```
ayushkovskiy: Please note that the method `upload_and_map_config` creates directory `/var/storage/nmrc`, so if a user has directory `storage://~/nmrc`, they will be in conflict. Let's maybe change `local_nmrc_folder` to `/var/storage/.nmrc`?

neuromation/cli/job.py
dalazx: should we handle edge cases such as same volumes etc? (prob out of scope)

neuromation/cli/job.py
dalazx: I think this is wrong as this is not corresponding to what `neuro ls storage:/` returns.

neuromation/cli/job.py
asvetlov: Sounds good

neuromation/cli/job.py
ayushkovskiy: looking at the code, I can say that this case is handled properly: the result collection `volumes` is of type `Set` (but in order to reduce number of loop iterations, I changed input array to the type of `Set` as well). 

However, this is a good case indeed, `neuro run --volume=ALL --volume=<custom>`, so since `--volume=ALL` already mounts *all possible* volumes, I changed the code to forbid this option together with any other option -- thank you for pointing this out!

neuromation/cli/job.py
ayushkovskiy: this is question of taste, let's ask our users (cc @mariyadavydova).

Currently:
- `--volume=HOME`: `storage://alice/` -> `/var/storage/home/`

We need to implement `--volume=ALL`. Suppose, `alice` is using this command. There are 2 options:

1. replace user's name with `home` in order to be consistent with `--volume=HOME` (and not cause surprises for our users):
- `storage://alice/dir1` -> `/var/alice/dir1`
- `storage://bob/dir2` -> `/var/bob/dir2`

2. as @dalazx proposes, do not change anything and thus simplify algorithm for `--volume=ALL`:
- `storage://alice/dir1` -> `/var/alice/dir1`
- `storage://bob/dir2` -> `/var/bob/dir2`

serhiy-storchaka: What if there is a user "home"?

dalazx: there won't be any surprises as it is `ALL` and not `HOME`. we are not changing `HOME` behavior.

I think we should indicate a user's home directory via an env var (e.g `$HOME`).

another option is to mount the tree as is, but add an additional mount (a sort of a link) for `home`.

ayushkovskiy: > What if there is a user "home"?

if we chose the single mount, this is fine (in both cases, either we rename username to `home` or not, it's not a problem). The problem will appear if we choose to mount both.

ayushkovskiy: > we are not changing HOME behavior.

true.

> there won't be any surprises as it is ALL and not HOME

let's wait for our user's opinion.

dalazx: @ayushkovskiy I am your user as well. I want to try to run a file server and do not want to mess around some weird directories such as `home`, but rather want to get the same hierarchy as our storage API exposes.

neuromation/cli/job.py
dalazx: the good thing is that we have the corresponding env var for this which allows us to change stuff.

neuromation/cli/job.py
ayushkovskiy: So do we agree on the following:
- ` --volume=HOME` mounts user's home directory to `/var/storage/home`, env var: `NEUROMATION_STORAGE_ROOT=/var/storage`
- `--volume=ALL` mounts everything available to `/var/storage` (including user's home dir mounted to `/var/storage/{username}`), env vars: `NEUROMATION_STORAGE_ROOT=/var/storage`, `NEUROMATION_STORAGE_HOME=/var/storage/{username}`

asvetlov: The home directory is mounted as `/var/storage/home` and available under `${NEUROMATION_STORAGE_ROOT}/home` if `--volume=HOME`.
At the same time it is mounted as `/var/storage/{username}` and available under `${NEUROMATION_STORAGE_HOME` if `--volume=ALL`.

Does anybody feel a little inconsistency?

dalazx: @asvetlov agree it might seem inconsistent.
`HOME` also mounts `"storage://neuromation/public:/var/storage/neuromation:ro"` which is even more inconsistent.
I think that we just need to distinguish the two variants properly.
`HOME` - happens to be a set of hardcoded mounts.
`ALL` - a tree similar to what storage API exposes.

re your concern, should we not use `$NEUROMATION_STORAGE_ROOT` for `HOME`?

I think the goal here is not to introduce any regressions at least.

asvetlov: Hard question.
Wild idea: we can keep existing mounts for backward compatibility *and* provide them in a consistent way under different path, e.g. `/var/neuro`. 
`/var/neuro` is a storage root, `/var/neuro/asvetlov` becomes my home etc. 
Thus `--volume=HOME` becomes a subset of `--volume=ALL` (plus legacy `/var/storage` mount which will be removed eventually, say in 6 months).

The downside is the absence of an ability to show deprecation warning in `neuro run` command, only in docs.

dalazx: brilliant!

ayushkovskiy: > Thus --volume=HOME becomes a subset of --volume=ALL (plus legacy /var/storage mount which will be removed eventually, say in 6 months).

@asvetlov do you mean with `--volume=ALL` we mount everything unchanged to `/var/neuro` (without mounting anything to `/var/storage`), and with `--volume=HOME` we mount all as-is?

> The downside is the absence of an ability to show deprecation warning in neuro run command, only in docs.

Why not? I don't see :(

asvetlov: With `--volume=HOME` we mount both `/var/storage/home` and `/var/neuro/asvetlov`, the same for public: `/var/storage/neuromation` and `/var/neuro/neuromation/public`.

dalazx: I think we can keep `HOME` as is, but address the inconsistency in `ALL` as @asvetlov suggested if we ever decide to drop `HOME`.

asvetlov: Yes, this is an option. If we decide to not touch HOME but remove it entirely somewhen we can add a deprecation now.
Is there a reason for mounting home and neuromation public only but skipping all other shared resources? Is `--volume=HOME` useful at all if we have `--volume=ALL` implemented?

dalazx: not sure. I guess we'll find out once we onboard more people. :)

asvetlov: Agree

ayushkovskiy: OK thanks everyone for sharing your opinions, here's the final version:
- for `--volume=HOME`, we don't change anything (just print a deprecation warning)
- for `--volume=ALL`, we mount everything available (by permissions) to `/var/neuro` and expose 2 variables: `NEUROMATION_ROOT=/var/neuro`, `NEUROMATION_HOME=/var/neuro/{username}`

we also forbid both `--volume=ALL` and any other `--volume` option.

mariyadavydova: I like the final version, thank you!

Considering HOME vs ALL: I think that the usage of these options is merely a question of examples we provide. Somehow I completely missed the HOME option being available. I assume that many existing users may be unaware of it because they are used to mounting particular URIs and PATHs. So we can actively promote ALL and never speak about HOME, and deprecate it soon.

I have a question though: from the technical point of view, if one have access to many shared resources, is it harder to mount ALL than to mount 2-3 paths? I mean can we safely recommend users to use ALL, or is there some trade off?

dalazx: In general I'd rather promote specific mounts rather than `ALL` to embrace isolation. though a user should be informed that there's an option to mount the whole available tree if needed.

I remember an interesting point @truskovskiyk brought when he was working on Hokutake. Kyryl did not even want `HOME` as he was afraid to corrupt the dataset and weights for the project by chance.

mariyadavydova: Indeed, that's a very good point. Corrupting the dataset or the weights may be extremely expensive accident.

All right, I will stick to recommending mounting specific URIs as a best practice in our tutorials and receipts, while providing a information about ALL shortcut for something like quick experiments or data exploration.

ayushkovskiy: > as he was afraid to corrupt the dataset and weights for the project by chance

How about we introduce (in future) an option or a syntax extension to not mount or mount in read-only mode some specific important directories? Though now it's hard for me to propose any such syntax

neuromation/cli/job.py
ayushkovskiy: not an issue anymore since we decided to use different mountpoints: /var/storage and /var/neuro

neuromation/cli/job.py
mariyadavydova: @ayushkovskiy This is a very good idea! We can introduce something like this:

```
> neuro storage protect storage:project/data
> neuro run ... --volume storage:project/data:/var/storage/data:rw ...
Warning: You are trying to mount storage:project/data with read-write access, while this directory is marked as protected. Access us downgraded to read-only. To get read-write access for this directory please remove the protection mode with
neuro storage protect --remove storage:project/data
```

dalazx: looks too complicated in my opinion.

mariyadavydova: I think it should be an option for "serious teams, concerned about data security". I mean, in most cases you can simply watch out your commands and use proper access modifiers. But if you are **really** concerned, you may additionally use such an option as a safety net. This may be especially relevant for teams.

ayushkovskiy: I agree we should print a warning for all `rw` volumes

neuromation/cli/job.py
```diff
@@ -168,8 +183,8 @@ def job() -> None:
     multiple=True,
     help="Mounts directory from vault into container. "
     "Use multiple options to mount more than one volume. "
-    "--volume=HOME is an alias for storage://~:/var/storage/home:rw and "
-    "storage://neuromation/public:/var/storage/neuromation:ro",
+    f"--volume=HOME is an alias for storage://~:{STORAGE_MOUNTPOINT}/home:rw and "
```
asvetlov: String fomatting operations in docstrings decrease the code readability, not increase it.

ayushkovskiy: fixed

neuromation/cli/job.py
ayushkovskiy: https://github.com/neuromation/platform-client-python/issues/1018

neuromation/cli/job.py
```diff
@@ -176,8 +191,8 @@ def job() -> None:
     multiple=True,
     help="Mounts directory from vault into container. "
     "Use multiple options to mount more than one volume. "
-    "--volume=HOME is an alias for storage://~:/var/storage/home:rw and "
-    "storage://neuromation/public:/var/storage/neuromation:ro",
+    f"--volume=HOME is an alias for storage://~:/var/storage/home:rw and "
```
asvetlov: Very minor: `f` is redundant here and below.

ayushkovskiy: oops :upside_down_face:

PR #981, opened 2019-08-23
Concurrent upload/download.

serhiy-storchaka:
Concurrently upload/download files and directories using the traditional protocol.

PR #984, opened 2019-08-23
Concurrent upload/download.

serhiy-storchaka:
Concurrently upload/download files and directories using the traditional protocol.

Recreation of #981.


serhiy-storchaka:
There is still minor problem with progress indication. When transfer a directory with a large number of files you can get a mess.

asvetlov:
I have a design note for all concurrent download/upload implementation, not this particular PR only.
On master now we call progress callbacks from the same task as was used for calling API, e.g. `await client.storage.upload_dir()`.
In this PR child task is used.
There is the only subtle difference from the caller perspective: modification of context var performed from callback is not visible from the caller task because spawn task has own context copy.
Not sure how it is important but I can see the point of confusion.

The fix can use async queue to call all report callbacks from root task only.

serhiy-storchaka:
Although large files are transferred using a single stream.

For example, 1 GiB file with random content:
```sh
dd if=/dev/urandom of=1G bs=1M count=1K
```
`neuro storage cp 1G storage:`
concurrent: 1 minute 30 seconds
websockets: 1 minute 20 seconds

`neuro storage cp storage:1G /tmp`
concurrent: 8 minutes 40 seconds
websockets: 40 seconds

serhiy-storchaka:
Yet example, directory containing 1000 tiny files:
```sh
mkdir 1000files
for i in `seq 1000`; do echo $i >1000files/$i.txt; done
```

`neuro storage cp -r 1000files storage:`
master: 3 minutes 50 seconds
concurrent: 57 seconds
websockets: 21 seconds
load: 46 seconds

`neuro storage cp -r storage:1000files /tmp`
master: 4 minutes 20 seconds
concurrent: 57 seconds
websockets: 15 seconds
load: 38 seconds

serhiy-storchaka:
I was not able to test downloading 1 GiB file by `neuro storage load` because it failed with errors like "404 page not found" or "AWS copying failed with code 1".

neuromation/api/storage.py
```diff
@@ -337,14 +342,22 @@ def _uri_to_path(self, uri: URL) -> str:
...
+        async with self._file_sem:
+            folder = sorted(
+                src_path.iterdir(), key=lambda item: (item.is_dir(), item.name)
+            )
         for child in folder:
             name = child.name
             if child.is_file():
-                await self._upload_file(src_path / name, dst / name, progress=progress)
+                tasks.append(
+                    self._upload_file(src_path / name, dst / name, progress=progress)
```
asvetlov: Small note: please create tasks explicitly, e.g.
```
                tasks.append(loop.create_task(
                    self._upload_file(src_path / name, dst / name, progress=progress)
                 ))
```
1. `tasks` variable contains task objects after the change, not coroutines.
2. In my experince, I've found that asyncio tricks for implicit wrapping coroutines in tasks by `ensure_future()` and `gather()` is confusing.

neuromation/api/storage.py
```diff
@@ -357,6 +370,7 @@ def _uri_to_path(self, uri: URL) -> str:
                         f"Cannot upload {child}, not regular file/directory",
                     )
                 )  # pragma: no cover
+        await asyncio.gather(*tasks)
```
asvetlov: Here there is asyncio quirk: `gather` doesn't cancel inner tasks if a future returned by `gather` call is canceled.

As you don't use results returned by `await gather()` the `gather()` is sub-optimal here I think.
Please consider the following instead:
```
try:
    done, tasks = await asyncio.wait(tasks, return_when=asyncio.FIRST_EXCEPTION)
    for task in done:
        await task
except BaseException:
    for task in tasks:
        task.cancel()
    await asyncio.wait(tasks)  # wait for actual cancellation, ignore all exceptions raised from tasks
```

What do you think about?

asvetlov:
Suggestion: please merge the master into the PR branch.
I may make the test suite more stable.

neuromation/api/storage.py
serhiy-storchaka: Since I use a helper function to avoid a code duplication, I prefer to keep this list containing coroutines and convert them to tasks in the helper function. To avoid repetitions.

neuromation/api/storage.py
serhiy-storchaka: Good catch, thanks!

This quirk is even specially highlighted in the documentation.

neuromation/api/storage.py
asvetlov: Yes, I see. Good choice.

Issue #982, opened 2019-08-23
Add TPU support for `neuro run/submit`


dalazx:
The job payload changes were implemented here:https://github.com/neuromation/platform-api/pull/844/files

Also going to patch this part
https://github.com/neuromation/platform-api/blob/c45a965f1319a2f798a46d5c2d3e346ceb7792e7/platform_api/api.py#L63
to add an optional `"tpu": {"type": "v3-8", "software_version": "1.14"}`

neuro submit should expose two new options: `--tpu-type` and `--tpu-sw-version`.

shagren:
closed via #985

PR #985, opened 2019-08-23
TPU feature implementation

shagren:
TPU Implementation  #982

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,8 +259,8 @@ Name | Description|
 |----|------------|
 |_\-g, --gpu NUMBER_|Number of GPUs to request  \[default: 0]|
 |_\--gpu-model MODEL_|GPU to use  \[default: nvidia\-tesla-k80]|
-|_\--tpu-type TYPE_|TPU to use|
-|_\--tpu-sw-version VERSION_|Requested TPU version|
+|_\--tpu-type TYPE_|TPU type to use|
+|_\--tpu-sw-version VERSION_|Requested TPU software version|
 |_\-c, --cpu NUMBER_|Number of CPUs to request  \[default: 0.1]|
 |_\-m, --memory AMOUNT_|Memory amount to request  \[default: 1G]|
 |_\-x, --extshm / -X, --no-extshm_|Request extended '/dev/shm' space  \[default: True]|
@@ -1353,8 +1353,8 @@ Name | Description|
 |----|------------|
 |_\-g, --gpu NUMBER_|Number of GPUs to request  \[default: 0]|
 |_\--gpu-model MODEL_|GPU to use  \[default: nvidia\-tesla-k80]|
-|_\--tpu-type TYPE_|TPU to use|
-|_\--tpu-sw-version VERSION_|Requested TPU version|
+|_\--tpu-type TYPE_|TPU type to use|
+|_\--tpu-sw-version VERSION_|Requested TPU software version|
 |_\-c, --cpu NUMBER_|Number of CPUs to request  \[default: 0.1]|
 |_\-m, --memory AMOUNT_|Memory amount to request  \[default: 1G]|
 |_\-x, --extshm / -X, --no-extshm_|Request extended '/dev/shm' space  \[default: True]|
--- a/README.md
+++ b/README.md
@@ -259,6 +259,8 @@ Name | Description|
 |----|------------|
 |_\-g, --gpu NUMBER_|Number of GPUs to request  \[default: 0]|
 |_\--gpu-model MODEL_|GPU to use  \[default: nvidia\-tesla-k80]|
+|_\--tpu-type TYPE_|TPU to use|
+|_\--tpu-sw-version VERSION_|Requested TPU version|
 |_\-c, --cpu NUMBER_|Number of CPUs to request  \[default: 0.1]|
 |_\-m, --memory AMOUNT_|Memory amount to request  \[default: 1G]|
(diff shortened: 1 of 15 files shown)


neuromation/api/config_factory.py
```diff
@@ -257,6 +257,8 @@ def _serialize_resource_preset(
             "memory_mb": resource_preset.memory_mb,
             "gpu": resource_preset.gpu,
             "gpu_model": resource_preset.gpu_model,
+            "tpu_type": resource_preset.tpu_type,
```
shagren: Let's store in plain lines

neuromation/api/jobs.py
```diff
@@ -414,27 +421,42 @@ def _raise_for_invalid_commit_chunk(obj: Dict[str, Any], expect_started: bool) -
 
 
 def _resources_to_api(resources: Resources) -> Dict[str, Any]:
-    value = {
+    value: Dict[str, Any] = {
```
shagren: Typing

neuromation/cli/formatters/config.py
```diff
@@ -1,35 +1,66 @@
 from sys import platform
+from typing import Dict
 
 from click import style
+from tabulate import tabulate
```
shagren: It helps format tables with mininal width

neuromation/cli/formatters/config.py
```diff
@@ -1,35 +1,66 @@
...
+            yes, no = "✔︎", "✖︎"
+        has_tpu = False
+        for preset in presets.values():
+            if preset.tpu_type:
+                has_tpu = True
+                break
+
+        table = []
+        headers = ["Name", "#CPU", "Memory", "Preemptible", "GPU"]
+        if has_tpu:
```
shagren: Show TPU column if having presets with tpu

neuromation/cli/formatters/config.py
```diff
@@ -1,35 +1,66 @@
...
+            headers.append("TPU")
+
+        for name, preset in presets.items():
+            gpu = ""
+            if preset.gpu:
+                gpu = f"{preset.gpu} x {preset.gpu_model}"
```
shagren: Show GPU in one column instead two. It will save few spaces in table

neuromation/cli/formatters/config.py
```diff
@@ -1,35 +1,66 @@
...
+                preset.cpu,
+                preset.memory_mb,
+                yes if preset.is_preemptible else no,
+                gpu,
+            ]
+            if preset.tpu_type:
+                tpu = f"{preset.tpu_type}/{preset.tpu_software_version}"
+                row.append(tpu)
+            table.append(row)
+        return tabulate(  # type: ignore
```
shagren: ignore types because typeshed collection having too old version :(

neuromation/cli/formatters/config.py
```diff
@@ -1,35 +1,66 @@
...
-                ).rstrip()
-            )
+        indent: str = "  "
         return (
             style("User Configuration", bold=True)
             + ":\n"
             + indent
             + f"\n{indent}".join(lines)
+            + "\n"
+            + self._format_presets(root.resource_presets)
```
shagren: indent will be removed, in other case presets table with TPU column can be wider than 80 chars

README.md
```diff
@@ -259,6 +259,8 @@ Name | Description|
 |----|------------|
 |_\-g, --gpu NUMBER_|Number of GPUs to request  \[default: 0]|
 |_\--gpu-model MODEL_|GPU to use  \[default: nvidia\-tesla-k80]|
+|_\--tpu-type TYPE_|TPU to use|
```
dalazx:
```suggestion
|_\--tpu-type TYPE_|TPU type to use|
```

README.md
```diff
@@ -259,6 +259,8 @@ Name | Description|
...
+|_\--tpu-sw-version VERSION_|Requested TPU version|
```
dalazx:
```suggestion
|_\--tpu-sw-version VERSION_|Requested TPU software version|
```

shagren:
> You have plain `tpu_type` / `tpu_software_version` in `RunPreset` but nested `tpu` dataclass in `Resources`.
> Is there reason why?

Because I would like to use plain serialization for yaml. And would like similar structure between yaml and dataclass

asvetlov:
So use a plain `Resources` type as well.
Nested classes are not bad at all, but we have `gpu`/`gpu_model` already.

neuromation/cli/formatters/config.py
asvetlov: The only thing I dislike is that `tabulate` installs an executable command as a side effect.
Thus, `pip install neuromation` actually installs `neuro`, `docker-credential-neuro` and `tabulate` console scripts.
Not sure how is it harm, though.

shagren: OK, I will try to avoid binary install

shagren:
@asvetlov, fixed

setup.py
```diff
@@ -37,6 +37,7 @@
         "humanize>=0.5",
         "psutil>=5.5.0",
         "typing_extensions>=3.7.2",
+        "tabulate>=0.8.3",
```
asvetlov: Please duplicate `tabulate==0.8.3` (strong form) in `requiremens/base.txt`.
It allows dependabot to check if everything is still going smoothly after future `tabulate` releases.

neuromation/cli/formatters/config.py
shagren: No way (

Issue #988, opened 2019-08-26
Support job operations via job-URI


ayushkovskiy:
Patch `resolve_job()` to accept URI and use its `host` part as filter while resolving the job-name.

PR #992, opened 2019-08-26
Support job operations via job-URI

ayushkovskiy:
Closes https://github.com/neuromation/platform-client-python/issues/988


neuromation/api/url_utils.py
```diff
@@ -73,6 +74,7 @@ def _normalize_uri(uri: URL, username: str) -> URL:
         if path or uri.host:
             uri = uri.with_path(path)
 
+    assert uri.host, f"Could not resolve resource owner for URI {resource}"
```
ayushkovskiy: AFAIK, `uri.host is None` is impossible. Though, I could not find fully covering unit tests on `_normalize_uri` -- should I create an issue?

neuromation/cli/utils.py
```diff
@@ -436,13 +436,33 @@ def volume_to_verbose_str(volume: Volume) -> str:
...
 
 
-async def resolve_job(client: Client, id_or_name: str) -> str:
+async def resolve_job(
+    id_or_name_or_uri: str, *, client: Client, default_user: str
+) -> str:
+    if id_or_name_or_uri.startswith("job:"):
+        uri = _normalize_uri(id_or_name_or_uri, username=default_user)
+        id_or_name = uri.path.lstrip("/")
+        assert uri.host
```
ayushkovskiy: this assert is redundant (since same check is present in `_normalize_uri`) and is equivalent to `# type: ignore`.

neuromation/cli/utils.py
```diff
@@ -466,6 +486,12 @@ def parse_resource_for_sharing(uri: str, root: Root) -> URL:
     return uri_from_cli(uri, root.username, allowed_schemes=("storage", "image", "job"))
 
 
+async def normalize_job_resource(uri: URL, root: Root) -> URL:
+    job_id = await resolve_job(str(uri), client=root.client, default_user=root.username)
+    owner = root.username if not uri.host or uri.host == "~" else uri.host
```
ayushkovskiy: AFAIK, we are going to drop tilde-expansion of the username, so let's not extract this logic to a separate method (see same logic in `_parse_as_neuro_image`)

neuromation/cli/utils.py
```diff
@@ -436,13 +436,32 @@ def volume_to_verbose_str(volume: Volume) -> str:
...
     try:
-        jobs = await client.jobs.list(name=id_or_name)
+        jobs = await client.jobs.list(name=id_or_name, owners={owner})
     except Exception as e:
         log.error(f"Failed to resolve job-name '{id_or_name}' to a job-ID: {e}")
     if jobs:
+        if len(jobs) > 1:
+            log.debug(
+                f"Found {len(jobs)} jobs matching name={id_or_name} owner={owner}: "
+                " ,".join(job.id for job in jobs)
```
serhiy-storchaka:
```suggestion
                ", ".join(job.id for job in jobs)
```

neuromation/cli/utils.py
serhiy-storchaka: `job:/foo` is not the same as `job:foo`. The latter refers to job foo of the current user. The former refers to... all jobs of user foo?

neuromation/cli/utils.py
ayushkovskiy: thankx

ayushkovskiy:
> job:/foo is not the same as job:foo. The latter refers to job foo of the current user. The former refers to... all jobs of user foo?

Yes I mentioned this. However, currently we parse images wrongly, see tests `test_parse_as_neuro_image_with_scheme_1_slash_*`. So let's keep the logic `job:/foo == job:foo` in this PR, but refactor it later. I'll create an issue.

ayushkovskiy:
> Yes I mentioned this. However, currently we parse images wrongly, see tests test_parse_as_neuro_image_with_scheme_1_slash_*. So let's keep the logic job:/foo == job:foo in this PR, but refactor it later. I'll create an issue.

Actually, for both jobs and images, `scheme:/name` is same as `scheme:name` because currently we don't differentiate between absolute and relative image/job name. Thus this is not an issue.

neuromation/cli/utils.py
```diff
@@ -485,12 +485,6 @@ def parse_resource_for_sharing(uri: str, root: Root) -> URL:
     return uri_from_cli(uri, root.username, allowed_schemes=("storage", "image", "job"))
 
 
-async def normalize_job_resource(uri: URL, root: Root) -> URL:
```
ayushkovskiy: unused method

neuromation/cli/utils.py
```diff
@@ -436,13 +436,36 @@ def volume_to_verbose_str(volume: Volume) -> str:
     )
 
 
-async def resolve_job(client: Client, id_or_name: str) -> str:
+async def resolve_job(
+    id_or_name_or_uri: str, *, client: Client, default_user: str
```
asvetlov: I'm curious why the function signature is changed at all?

neuromation/api/url_utils.py
```diff
@@ -51,7 +51,8 @@ def normalize_storage_path_uri(uri: URL, username: str) -> URL:
     return _normalize_uri(uri, username)
 
 
-def _normalize_uri(uri: URL, username: str) -> URL:
+def _normalize_uri(resource: Union[URL, str], username: str) -> URL:
```
asvetlov: Please keep the function untouched but call `_normalize_uri(URL(s), username)` where needed.
Union is not a perfect argument type, here it can be avoided easily.

PR #993, opened 2019-08-27
Replace storage_path str with storage_uri URL in Volume datastructure

asvetlov:


neuromation/cli/job.py
```diff
@@ -898,7 +898,7 @@ def format_fail(job: str, reason: Exception) -> str:
     return (
         local_nmrc_path,
         Volume(
-            storage_path=storage_nmrc_folder,
+            storage_uri=URL(storage_nmrc_folder),
```
serhiy-storchaka: I would make the `storage_nmrc_folder` variable of type URL (line 885).

asvetlov: Nice idea! Done

Issue #989, opened 2019-08-26
Support `neuro acl` operations with job-name in job-URI


yartem:
Patch `parse_resource_for_sharing` to resolve `job://owner/job-name` -> `job://owner/job-ID` and then share it (see https://github.com/neuromation/platform-client-python/issues/968)

dalazx:
URIs should not have names in them because names are not unique.

ayushkovskiy:
> URIs should not have names in them because names are not unique.

why not? we have a defined algorithm of resolving job by job name, and by introducing job URIs we add the users ability to address other users' jobs, so job names are resolved to job IDs in the same way.

ayushkovskiy:
note that job URIs like `job://{owner}/{job-name}` are faced only to the user: the job-name is resolved to job-ID on CLI-side, so the server always gets unambigous URIs.

serhiy-storchaka:
You can now share job by name. https://github.com/neuromation/platform-api/issues/1221

It shares all past, running and future jobs with this name.

Issue #967, opened 2019-08-20
Complete support job-URI


yartem:
when a job was shared, the user expects to address it via URI, however only bare job-id (or job-name) works now:
```
$ n status job://foo/job-35f32a08-7877-4ae9-bd04-5ffcebc38e15
ERROR: Failed to resolve job-name 'job://foo/job-35f32a08-7877-4ae9-bd04-5ffcebc38e15' to a job-ID: {"error": "{0: DataError(value should be None), 1: DataError(String is longer than 40 characters)}"}
ERROR: 404: Not Found
$ n status job-35f32a08-7877-4ae9-bd04-5ffcebc38e15
[OK]
```

ayushkovskiy:
*Problem*: even though we operate with job-URIs server-side, we don't support them client-side.

*Use-case*:
- alice has job named `test-job` running,
- bob has job named `test-job` running,
- bob shares his job with alice,
- alice sees two jobs with name `test-job`, but her client automatically refers to only one of them (latest started). 

*Plan*:
Thus, we need to do the following things to fully support job URI:
1. patch api method `Jobs.list()` to send filter by owner (filtering by owner is already supported server-side)
2. patch `resolve_job()` to accept URI and use its `host` part as filter while resolving the job-name.
3. patch `parse_resource_for_sharing` to resolve `job://owner/job-name` -> `job://owner/job-ID` (see https://github.com/neuromation/platform-client-python/issues/968)
4. print job's URI somewhere in `neuro status`: otherwise it's unclear how to use them (see https://github.com/neuromation/platform-client-python/issues/968#issuecomment-524585629)

PR #999, opened 2019-08-27
Support `neuro acl` operations with job-name in job-URI

ayushkovskiy:
Closes https://github.com/neuromation/platform-client-python/issues/989


neuromation/cli/utils.py
```diff
@@ -489,6 +489,25 @@ def parse_resource_for_sharing(uri: str, root: Root) -> URL:
...
+async def parse_and_resolve_resource_for_sharing(uri: str, root: Root) -> URL:
+    """ Parses the neuromation resource URI string + resolves job-name to job-ID.
+    Available schemes: storage, image, job. For image URIs, tags are not allowed.
+    """
+    uri_obj = parse_resource_for_sharing(uri, root)
+    if uri_obj.scheme == "job":
+        job_id = await resolve_job(
+            str(uri_obj), client=root.client, default_user=root.username
+        )
+        uri_obj = uri_obj.with_path(f"/{job_id}")
```
asvetlov:
```suggestion
        uri_obj = uri_obj.with_path(job_id)
```

neuromation/cli/utils.py
```diff
@@ -489,6 +489,25 @@ def parse_resource_for_sharing(uri: str, root: Root) -> URL:
     return uri_from_cli(uri, root.username, allowed_schemes=("storage", "image", "job"))
 
 
+async def parse_and_resolve_resource_for_sharing(uri: str, root: Root) -> URL:
+    """ Parses the neuromation resource URI string + resolves job-name to job-ID.
+    Available schemes: storage, image, job. For image URIs, tags are not allowed.
+    """
+    uri_obj = parse_resource_for_sharing(uri, root)
```
serhiy-storchaka: `parse_resource_for_sharing()` is only used here. I suggest to inline it.

neuromation/cli/utils.py
```diff
@@ -489,6 +489,25 @@ def parse_resource_for_sharing(uri: str, root: Root) -> URL:
...
+    if uri_obj.scheme == "job":
+        job_id = await resolve_job(
+            str(uri_obj), client=root.client, default_user=root.username
+        )
+        uri_obj = uri_obj.with_path(job_id)
+    return uri_obj
+
+
+async def normalize_job_resource(uri: URL, root: Root) -> URL:
```
serhiy-storchaka: Is it used anywhere?

tests/cli/test_utils.py
```diff
@@ -403,6 +405,66 @@ def _job_entry(job_id: str) -> Dict[str, Any]:
...
+    owner_name = "owner-name"
+    uri = f"job://{owner_name}/{job_name}"
+    job_id = "job-id-1"
+    JSON = {"jobs": [_job_entry(job_id)]}
+
+    async def handler(request: web.Request) -> web.Response:
+        # Since `resolve_job` excepts any Exception, `assert` will be caught there
+        name = request.query.get("name")
+        if name != job_name:
+            pytest.fail(f"received: {name}")
```
serhiy-storchaka:
```
            pytest.fail(f"received name: {name}")
```
or
```
            pytest.fail(f"received: {request.query}")
```
or even
```
        assert request.query.get("name") == job_name
```

neuromation/cli/utils.py
ayushkovskiy: let's keep it as-is for now: since it's non-async function that encapsulates quite large logic of uri parsing, which is already covered with unit tests, I decided to simply wrap it with an async function `parse_and_resolve_resource_for_sharing` rather than replace it with another function. Let's keep this refactoring for the future (if needed) in order for this PR to remain smaller.

neuromation/cli/utils.py
ayushkovskiy: you're right, it's redundant method left from older refactorings! Thank

tests/cli/test_utils.py
ayushkovskiy: > assert request.query.get("name") == job_name

this doesn't work as this assertion error will be caught by try-catch block within `resolve_job()`, so the result of test will be silently different. I agree, `except Exception` sounds like a bad design, but for this method, which follows the strategy "try to resolve, if *anything* goes wrong, roll back to the original argument silently", it's fine IMO.

ayushkovskiy: fixed to 
```
pytest.fail(f"received name: {name}")
```
thanks

neuromation/cli/utils.py
serhiy-storchaka: Okay.

ayushkovskiy:
> But in general, I think it should be fixed by replacing Exception with a tuple of expected exceptions. It silences also MemoryError and RecursionError.

@serhiy-storchaka What is the full list of exceptions advised not to catch? `MemoryError`, `RecursionError`, `CancelledError`, something else?

serhiy-storchaka:
I think it is better to have a list of exceptions which should be caught rather of the list of exceptions which should not be caught. But this is a separate issue.

ayushkovskiy:
@serhiy-storchaka 's comment on excepting too broad class of exception is considered.

neuromation/cli/utils.py
```diff
@@ -455,6 +456,8 @@ def volume_to_verbose_str(volume: Volume) -> str:
     details = f"name={id_or_name}, owner={owner}"
     try:
         jobs = await client.jobs.list(name=id_or_name, owners={owner})
+    except (MemoryError, RecursionError, CancelledError, AssertionError):
```
ayushkovskiy: @asvetlov please take a look at this change: do we miss some important exception that needs to be re-risen?

ayushkovskiy:
Very long discussion: in \#platform-code: https://neuromation.slack.com/archives/CE76BT03G/p1567434773079700

<details><summary>Click here to see full discussion</summary>
<p>

Aleksandr Danshyn Sep 2nd at 17:32
@here guys, I’d like to discuss the functionality we’ve recently merged https://github.com/neuromation/platform-client-python/pull/992
as far as I understand, our job URIs became ambiguous because the last path component may be either an ID or a name. and without neuro one won’t be able to distinguish the two cases (although we know that job IDs start with job- and are quite long, it may or may not be the case later on). moreover, a job URI with a name instead of an ID can lead to completely different jobs depending on the time.
curious, whether those are still URIs.
won’t that be confusing? what do you think?
ayushkovskiyayushkovskiy
#992 Support job operations via job-URI
Closes #988
Labels
CLIents, platform
Comments
6
<https://github.com/neuromation/platform-client-python|neuromation/platform-client-python>neuromation/platform-client-python | Aug 26th | Added by GitHub
37 replies

Mariya Davydova  3 days ago
Thank you for raising this question! I'm reading the related issues and PRs and will be back with some opinion later.

Serhiy Storchaka:house_with_garden:  3 days ago
See also https://github.com/neuromation/platform-client-python/pull/999
ayushkovskiyayushkovskiy
#999 Support `neuro acl` operations with job-name in job-URI
Closes #989
Labels
CLIents, platform
Comments
3
<https://github.com/neuromation/platform-client-python|neuromation/platform-client-python>neuromation/platform-client-python | Aug 27th | Added by GitHub

Serhiy Storchaka:house_with_garden:  3 days ago
We already supported resolving job:job-id -> job://myname/job-id. This PR will add also resolving job:job-name -> job://myname/job-id and job://otheruser/job-name -> job://otheruser/job-id.

Andrew Svetlov  3 days ago
Support job:job-{uuid} everywhere is a good thing without any doubt.
The URL is unique and not ambiguous.
I consider job:job-name as an alias for job:job-{uuid} but more readable and easier to remember.
So far so good.
Today @oleksandr.danshyn expressed an opinion that job names require own namespace (scheme), e.g. job-name:name.
Makes sense, I like it. job: is for unique permanent IDs, job-name: is for names that can be stollen by newer jobs.

Serhiy Storchaka:house_with_garden:  3 days ago
The only drawback of this feature is that if the user make an error in the job name, it will add a record about job://myname/wrong-job-name in the ACL database. The same may happen in the case of a temporary glitch on the server, on the client or between, because any error in resolving a job name is interpreted as "it is a job ID".

Andrew Svetlov  3 days ago
IMHO ACL database should not save names but IDs only because the name can be rebound to new job.

Serhiy Storchaka:house_with_garden:  3 days ago
Yes. The problem is that if we fail to resolve a job name (because it is actually a job ID or because of error) it will be interpreted as a job ID.

Artem Yushkovskiy  3 days ago
1. I think we should support job names for all operations, and allow user to completely avoid using job-ID (because it's ugly). Also, we don't implement algorithm detecting whether neuro status {value} uses a job-name or a job-ID in {value} by its structure (i.e., if it matches pattern job-{uuid}, then it's likely a job-ID) -- this is too complicated and thus unpredictable. Instead, we use higher precedence of job-name over job-id (for example, if a user has two jobs one with job-id job-ID and the other one with job-name job-ID, then neuro kill job-ID will kill the first one. This is how docker works as well.
2. Currently, we use clearly defined algorithm for resolving job-name to job-ID (current implementation: take a look at all  available jobs with name "job-name" and take the last one started). The aforementioned two PRs systematize the work with job URIs: now, the user can also specify the owner of the job, say: job:job-name == job://my-name/job-name, or job://other-name/job-name. If we see this syntax too complicated to use, we need rather to re-work the CLI operating this syntax, but not support it partially as we do now (for example, by supporting links to jobs as job-name and not supporting job://alice/job-name`). I see job's URI is well-defined by two set of rules: (1) same rules applied to all URIs (scheme://owner/resource-name`) and (2) our algorithm of resolving job-name to job-id.
3. I think, that these rules described above are strict enough for referring job as a resource. If not, then we can say that URI storage://artem/path/to is ambiguous as well, if we stick to the content (meaning) of the resource (in this exapmle, it's the content of the directory /path/to stored on storage, which, similary to job, is dynamic: we can remove replace the directory with another one while preserving the permissions).
So I think, we don't need to introduce another scheme job-name://... as job://... is already non-ambiguous up to the job-name resolution algorithm.
Context: please see the original issue of supporting job-URIs (partially closed): https://github.com/neuromation/platform-client-python/issues/967 (this discussion should have located there).

ayushkovskiyayushkovskiy
#967 Complete support job-URI
when a job was shared, the user expects to address it via URI, however only bare job-id (or job-name) works now:
$ n status job://foo/job-35f32a08-7877-4ae9-bd04-5ffcebc38e15
ERROR: Failed to resolve job-name 'job://foo/job-35f32a08-7877-4ae9-bd04-5ffcebc38e15' to a job-ID: {"error": "{0: DataError(value should be None), 1: DataError(String is longer than 40 characters)}"}
ERROR: 404: Not Found
$ n status job-35f32a08-7877-4ae9-bd04-5ffcebc38e15
[OK]
Show more
Labels
CLIents, UX, platform
Comments
1
<https://github.com/neuromation/platform-client-python|neuromation/platform-client-python>neuromation/platform-client-python | Aug 20th | Added by GitHub

Aleksandr Danshyn  3 days ago
thanks @artem.yushkovskiy
1. I think we all agree and it seems that this is what neuro currently does;
2. owner makes total sense: one either uses the relative form job:ID or the absolute one job://user/ID similarly to the rest of the schemes; not sure if I agree with (2);
3. resources can be changed/expanded of course, but it does not mean that you need to generate new URIs for any small change, but rather need to be able to unambiguously identify the resource; in your example with storage, the directory is unambiguously identified regardless of the content; the content is identified as subresources with their corresponding URIs; what you are proposing is to completely change the meaning of the path component with no easy way to identify the underlying resource; now we have two meanings, tomorrow you will want 5 meanings; this does not make much sense to me.

Artem Yushkovskiy  3 days ago
side idea: how about we introduce a flag --no-resolve-job-name (with a shorter name of course) that interprets job-name as job-ID (e.g. in case when one needs to share this exactly job)? For example, neuro kill --no-resolve-job-name job-id similar to neuro kill --no-resolve-job-name job://owner/job-id.
resources can be changed/expanded of course, but it does not mean that you need to generate new URIs for any small change
First, we need to define what it means, "the reource to change" for all the resources (image: new tag pushed?, storage: directory moved or any file within it changed? job: its status changed?). Because from the user perspective, I believe, two jobs with the same name (read as: two experiments runs) are the same as two images with different tags, meaning different versions of the same experiment. Thus, a scheme like job://owner/job-name is perfectly valid.
Second, we don't generate new URIs: on server-side we operate only with job URIs of the normalized form job://owner/job-ID (job-name is resolved to job-ID locally by the client and using the information specified by the user in job-URI).

Aleksandr Danshyn  3 days ago
all our commands already work with job IDs and names. we taught them to also work with URIs. why should we introduce a yet another confusing option with URIs that have a different meaning.
it looks like the only concern is the share thing. we all agreed that share is not ideal. we can’t we stick to unambiguous URIs for now and better make share smarter by making sure resources exist etc?

Aleksandr Danshyn  3 days ago
neuro share $(neuro status -o URI $NAME) alice read (edited)
:astonished:
1


Serhiy Storchaka:house_with_garden:  3 days ago
all our commands already work with job IDs and names
But they work by name only with your jobs, isn't? To specify both the owner and the name as a single argument we need a special syntax. And this syntax is a URI job://owner/name.

Artem Yushkovskiy  3 days ago
and better make share smarter by making sure resources exist etc
That's why I merged job URIs with support of job-names for all operations except neuro share without opening the discussion publicly. Of course we can close this PR and require to specify job-ID only when sharing a resource. If we (pardon me, if our users) want, we can also deprecate this syntax everywhere, but I don't think it's what they want.
neuro share $(neuro status -o URI $NAME) alice read
not sure if it looks simpler than job://owner/job-name
To specify both the owner and the name as a single argument we need a special syntax. And this syntax is a URI job://owner/name
True, but, as far as I understand, @oleksandr.danshyn’s concern is that the user can accidentally share a wrong job using this syntax. However, if this concern is valid, then we should deprecate all operations with jobs by their name (for example, one can kill their job by name, even if they think of another job).
Overall, as I already stated before, I think the problem is not around job-URI -- everyone agrees that URIs are too complicated to be exposed to end users -- the problem is around resource URIs in general. But the good news is that, as far as I am informed, @Mariya Davydova and @astaff are working on unified framework that will simplify work with URIs.

Aleksandr Danshyn  3 days ago
@serhiy.storchaka not only with yours, afaik, but with all jobs shared with you

Serhiy Storchaka:house_with_garden:  3 days ago
But not if you have several shared jobs with the same name?

Aleksandr Danshyn  3 days ago
you’ll get the last one right

Serhiy Storchaka:house_with_garden:  3 days ago
So if somebody will share with you a just ran job while you type a command you will run a command with wrong task?

Aleksandr Danshyn  3 days ago
@artem.yushkovskiy my concern is that our job URIs are no longer uniform and unambiguous, thus they are no longer URIs it seems.

Serhiy Storchaka:house_with_garden:  3 days ago
The syntax looks ambiguous to me. Names are just aliases (or symlinks) to ids. The downside is that supporting this feature is not free, every command needs to try to resolve the job name, and threat is as the job id if fail to resolve it as a name. And in the case of share, when the wrong job id does not fail immediately, the operation can finish successfully but with wrong effect.
Maybe use different syntax for job names? job://owner?name or job://owner#name or job-name://owner/name? I afraid it will confuse users. :disappointed:

Aleksandr Danshyn  3 days ago
@serhiy.storchaka might be the case, yes. this is why, again, share is not ideas, rather rudimentary. we may want to think about implementing share requests so that a user would confirm before we actually add the resource to their set of permissions.

Aleksandr Danshyn  3 days ago
I also thought of job:?name=$NAME as a legit alternative.
job://owner?name=$NAME to limit to scope. (edited)

Andrew Svetlov  2 days ago
another scheme is better than utilizing query or fragment parts; they have a separate semantics.

Artem Yushkovskiy  2 days ago
So please let's agree on the our following strategy:
1. we deprecate job://owner/job-name for all commands and **do not** support them for neuro-acl commands at all
2. we deprecate job://owner/job-name for all commands and **change the syntax** to either job-name://owner/job-name or something else for all possible commands in neuro-cli
3. we leave everything as-is: one can do anything but neuro-acl with job://owner/job-name, but neuro-acl commands require to use Job-ID in URI
4. anything else?

Serhiy Storchaka:house_with_garden:  1 day ago
3 looks reasonable to me. All other commands fail hard in case of error in job name resolving anyway. Only neuro-acl commands can silently do wrong things.

Artem Yushkovskiy  1 day ago
they can (should) print the ID of shared resource (job-URI with job-ID)

Aleksandr Danshyn  1 day ago
imo (3) does not make much sense: all commands except acl had been supporting name resolution even before we patched stuff. so no gain. am I missing something?

Aleksandr Danshyn  1 day ago
yet another possible solution could be neuro job share $NAME alice read
:heart:
1


Artem Yushkovskiy  1 day ago
but this will take large modifications in CLI

Aleksandr Danshyn  1 day ago
why do you think so?

Artem Yushkovskiy  1 day ago
in order to preserve symmetry over our CLI commands, we might want to deprecate neuro acl grant/revoke in favour of neuro {resource} grant/revoke for other resources. Otherwise it looks as a hack: to share any resource except job use neuro acl grant, for jobs use neuro job share

Aleksandr Danshyn  1 day ago
not sure what symmetry you are talking about. definitely that’s not a hack. acl works with URIs only

Andrew Svetlov  1 day ago
neuro acl grant storage $NAME alice read -- require kind of resource for grant/revoke command.
This is not a problem.
I really don't like that any job-subsystem command requires at least two REST API calls: one for name resolving and another one is for actual call.
Looks like pessimization

Artem Yushkovskiy  1 day ago
OK here's the solution we came up to recently: We drop support for job://{owner}/{job-name} in all neuro commands, such that all job URIs are parsed as following: job://{owner}/{job-ID}. Neuro-acl comments accept job resources only in the following format: job:{job-id} or job://{owner}/{job-ID}.
Thus, all neuro commands will normalize job resource w.r.t. the following algorithm:
1. if the URI was passed, it is interpreted as job://{owner}/{job-ID}
2. if a bare string S (without schema job) was passed, it is assumed to be a job-name and resolved as following:
1) lookup all jobs with job name S
2) if there's no user's jobs (only shared job), take the last created among shared jobs.
3) if there are at user's jobs, take the last created among them (this will prevent sabotaging other user's work by a user who creates jobs with same job-names and shares them: https://github.com/neuromation/platform-client-python/issues/1024#issuecomment-527159315)

serhiy-storchakaserhiy-storchaka
Comment on #1024 Should rise an exception on ambiguous job-name
I think the calling user should have a preference, and no exception should be raised.
Otherwise I could sabotage your work by granting jobs with names that match your job names. It will make using a job name for addressing a job non-guaranteed.
<https://github.com/neuromation/platform-client-python|neuromation/platform-client-python>neuromation/platform-client-python | Sep 2nd | Added by GitHub
:fire:
1


Andrew Svetlov  24 hours ago
I would propose additional rule: job name cannot start with job:.
By providing this we can eliminate a need for lookup if ID is provided.
Tab completion is hard though: if a user has neuro kill jo+TAB the completer should fetch all (active) jobs and suggest job: along with john if a job named john is present.
Thus, autocompletion for job ids/names is slow in our case by design.  We can recommend avoiding to assign names started with j though -- for sake of performance :slightly_smiling_face:

Artem Yushkovskiy  24 hours ago
job name cannot start with job:
job name cannot contain semicolon. And, as stated in the previous message, job-URI can contain only job-ID in path component, so this rule is redundant IMO

Andrew Svetlov  23 hours ago
sorry, you are right. Cool!
:ok_hand:
1


</p>
</details>

ayushkovskiy:
Closing this PR, will apply changes we agreed with in the discussion (see above) in a separate PR.

PR #1006, opened 2019-08-28
Generalize job.list() defaults

asvetlov:
There is no reason for things like `Optional[str]` if code works equally for `None` and empty string.
Also, restricting the accepted type to `Optional[Set[str]]` is worse than more wider `Iterable[str]`. Again, empty iterable is enough mark for *not set* value, there is no need for handling `None` type.


asvetlov:
Sorry, I've merge this PR with red state.
Need this tiny change for my docs branch.

Issue #1024, opened 2019-09-02
Should rise an exception on ambiguous job-name


ayushkovskiy:
when job is addressed via URI (`neuro status job://owner/job-name`), this URI incorporates owner's name, so multiple running/pending jobs is impossible case. However, if the job is addressed via bare job name (`neuro status job-name`), this job-name may refer to a self user's job or other user's job shared with the first user. Thus job-name is ambiguous, we should raise an exception. Currently, we resolve it as the calling user (see https://github.com/neuromation/platform-client-python/issues/988, https://github.com/neuromation/platform-client-python/pull/999/).

ayushkovskiy:
and also, if the the bare job name is used (not uri), we should not resolve apply filter by owner when resolving

serhiy-storchaka:
I think the calling user should have a preference, and no exception should be raised.

Otherwise I could sabotage your work by granting jobs with names that match your job names. It will make using a job name for addressing a job non-guaranteed.

ayushkovskiy:
> Otherwise I could sabotage your work by granting jobs

Good point! thank you @serhiy-storchaka , I agree that CLI should include `owner=myself` when resolving both `job-name` and `job:job-name` to job-ID. I'd like to close this issue.

PR #1026, opened 2019-09-03
Hotfix: remove annoying warning during job-name to job-ID resolution

ayushkovskiy:
Closes https://github.com/neuromation/platform-client-python/issues/1025

### DIFF
--- a/neuromation/cli/utils.py
+++ b/neuromation/cli/utils.py
@@ -461,11 +461,6 @@ async def resolve_job(
             f"{details} to a job-ID: {e}"
         )
     if jobs:
-        if len(jobs) > 1:
-            log.warning(
-                f"Found {len(jobs)} jobs matching {details}: "
-                ", ".join(job.id for job in jobs)
-            )
         job_id = jobs[-1].id
         log.debug(f"Job name '{id_or_name}' resolved to job ID '{job_id}'")
     else:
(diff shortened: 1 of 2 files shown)


neuromation/cli/utils.py
```diff
@@ -461,11 +461,6 @@ def volume_to_verbose_str(volume: Volume) -> str:
             f"{details} to a job-ID: {e}"
         )
     if jobs:
-        if len(jobs) > 1:
-            log.warning(
-                f"Found {len(jobs)} jobs matching {details}: "
-                ", ".join(job.id for job in jobs)
-            )
         job_id = jobs[-1].id
```
serhiy-storchaka: Would not be better to return always *your* job if there is such in the found jobs and return the last job only if no of them are your?

serhiy-storchaka: Ah, the owner is always specified.

ayushkovskiy: Yes, as we agreed in https://github.com/neuromation/platform-client-python/issues/1024, `neuro kill job-name` will also pass the parameter `owner=myself` when resolving `job-name` to job-ID

asvetlov:
Please make at least on Circle CI job green

asvetlov:
Forgot to mention: please add changelog file.
The PR worth a release, I want to provide a text to users why the release is needed.

ayushkovskiy:
> Forgot to mention: please add changelog file.

Thank you! added.

PR #1041, opened 2019-09-09
Run listing a local directory in an executor.

serhiy-storchaka:
It may be an expensive operation for large dirs, and it should
not block concurrent IO.

### DIFF
--- a/neuromation/api/storage.py
+++ b/neuromation/api/storage.py
@@ -451,10 +451,9 @@ class Storage(metaclass=NoPublicConstructor):
         except (FileExistsError, neuromation.api.IllegalArgumentError):
             raise NotADirectoryError(errno.ENOTDIR, "Not a directory", str(dst))
         await queue.put((progress.enter, StorageProgressEnterDir(src, dst)))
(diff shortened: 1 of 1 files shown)


neuromation/api/storage.py
```diff
@@ -451,10 +451,9 @@ def _is_remote_modified(self, local: os.stat_result, remote: FileStatus) -> bool
         except (FileExistsError, neuromation.api.IllegalArgumentError):
             raise NotADirectoryError(errno.ENOTDIR, "Not a directory", str(dst))
         await queue.put((progress.enter, StorageProgressEnterDir(src, dst)))
+        loop = asyncio.get_event_loop()
         async with self._file_sem:
-            folder = sorted(
```
asvetlov: I'm curious why sorting is not needed anymore.

serhiy-storchaka: Because in any case tasks start concurrently. It is not predicable what operation will start or finish first.

asvetlov: got it

PR #1048, opened 2019-09-11
Delete redundant file `out`

ayushkovskiy:
oops :roll_eyes:


ayushkovskiy:
thanks to @shagren for pointing out on this problem

Issue #1043, opened 2019-09-09
Add `neuro project init` command


ayushkovskiy:
the new command group `neuro project` with the only command so far `neuro project init`

PR #1049, opened 2019-09-11
Add `neuro project init` command

ayushkovskiy:
Closes https://github.com/neuromation/platform-client-python/issues/1043

### DIFF
--- a/neuromation/cli/project.py
+++ b/neuromation/cli/project.py
@@ -1,0 +1,43 @@
+import logging
+
+from cookiecutter.main import cookiecutter
+
+from .root import Root
+from .utils import async_cmd, command, group
+
+
+log = logging.getLogger(__name__)
+
+
+@group()
+def project() -> None:
+    """
+    Project operations.
+    """
+
+
+@command()
+@async_cmd()
+async def init(root: Root) -> None:
+    """
+    Initialize an empty project.
+
+    Examples:
+
+    # Initializes a scaffolding for the new project with the recommended
+    # project structure:
+    #   .
+    #   ├── {project_name}     # directory with your code
+    #   ├── data               # directory with your dataset
+    #   ├── notebooks          # directory with your jupyter notebooks
+    #   ├── Makefile           # directory with your datasets
+    #   ├── apt.txt            # system-wide requirements (apt-get install ...)
+    #   ├── requirements.txt   # Python requirements (pip install ...)
+    #   ├── setup.py           # project configuration script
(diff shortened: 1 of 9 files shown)


neuromation/cli/project.py
```diff
@@ -0,0 +1,43 @@
...
+async def init(root: Root) -> None:
+    """
+    Initialize an empty project.
+
+    Examples:
+
+    # Initializes a scaffolding for the new project with the recommended
+    # project structure:
+    #   .
+    #   ├── {project_name}     # directory with your code
```
asvetlov: Please drop the project structure tree.
Otherwise, you must update the docstring after every cookiecutter template change, I doubt if it is realistically possible.

ayushkovskiy: Hmm.. As far as I know, cookiecutter project structure is pretty much stable (changes cause lots of problems since we will build our funcitonality basing on it). Personally, I find project structure in docstring very informative.

cc @mariyadavydova

tests/cli/test_project.py
```diff
@@ -0,0 +1,28 @@
+import subprocess
+from pathlib import Path
+
+from .conftest import working_directory
+
+
+def test_project_init(tmp_path: Path) -> None:
+    with working_directory(tmp_path):
```
asvetlov: Please drop the test.
If you want to test something -- a check that mocked `cookiecutter.main.cookiecutter()` is called with the proper argument should be enough.
The test is not unit but integration, it fails if github connection is unstable (how often we encounted this last year?)

Also, the test fails on the template update, e.g. adding a license, etc. The client project should not care about the generated project structure, it is a responsibility of the template project.
Add these tests there if you want.

neuromation/cli/project.py
asvetlov: Please just drop it. Period.

mariyadavydova: This structure may change several times until we reach some stable point. As this command is separated from the template itself, we will most probably forget to update this docstring.

Let's drop it and add a link to the template repo instead (which is perfectly valid, we are using an established open-source tool and can tell about that).

ayushkovskiy: OK

tests/cli/test_project.py
asvetlov: The note is still valid

ayushkovskiy: OK, left uncovered.

PR #1059, opened 2019-09-16
Return error code for job start failures even with --detach

### LINKED ISSUE #1046: Return code is 0 when job didn't start
When `neuro run` has the `--detach` parameter it always returns zero exit code, even in situations when job was unable to run (e.g., when Cluster ScaleUp failed).

I think `neuro run` should return a non-zero exit code in situations when it knows job has failed.

Example:
```
# neuro run --name jupyter --detach …
Job ID: job-33c220ec-7c82-427e-939a-940490be59ac Status: pending
Name: jupyter
Http URL: https://jupyter--spbrspo.jobs-dev.neu.ro
…
Status: pending Initializing
Status: pending Scheduling
Status: failed ClusterScaleUpFailed (Failed to scale up the cluster to get more resources)
# echo $?
0
```

anayden:
Closes #1046

### DIFF
--- a/neuromation/cli/job.py
+++ b/neuromation/cli/job.py
@@ -927,7 +927,7 @@ async def run_job(
         # Even if we detached, but the job has failed to start
         # (most common reason - no resources), the command fails
         if job.status == JobStatus.FAILED:
-            exit_code = 1
+            exit_code = 125
 
     if exit_code is not None:
         sys.exit(exit_code)
--- a/neuromation/cli/job.py
+++ b/neuromation/cli/job.py
@@ -906,8 +906,7 @@ async def run_job(
         job = await root.client.jobs.status(job.id)
         progress(job)
     progress.close()
-    # Nothing to browse unless job is running
-    if browse and job.status == JobStatus.RUNNING:
+    if browse and job.status != JobStatus.FAILED:
         await browse_job(root, job)
 
     exit_code = None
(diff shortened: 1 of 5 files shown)


neuromation/cli/job.py
```diff
@@ -919,8 +921,16 @@ def format_fail(job: str, reason: Exception) -> str:
...
         await _print_logs(root, job.id)
-        res = await root.client.jobs.status(job.id)
-        sys.exit(res.history.exit_code)
+        job = await root.client.jobs.status(job.id)
+        exit_code = job.history.exit_code
+    else:
+        # Even if we detached, but the job has failed to start
+        # (most common reason - no resources), the command fails
+        if job.status == JobStatus.FAILED:
+            exit_code = 1
```
asvetlov: I'm curious why you use `job.history.exit_code` for an attached job but `1` for detached one?

anayden: When detached job fails, it means it didn't even start. Hence there's no exit code. `job.history.exit_code` for such job is `None`

asvetlov: Got you.

Looks like docker uses `126` exit code for the case: https://docs.docker.com/engine/reference/run/#/exit-status
It is compliant to chroot exit codes table.
Maybe we can return `126` as well?

asvetlov: Or `125`, I have no strong opinion here.

anayden: I like 125, will update now.

anayden:
A run of python 3.6 tests was successful. The tests are very flaky, but it's proven to work :) Merging.

PR #1069, opened 2019-09-24
Make the output of presets memory size more human readable.

### LINKED ISSUE #1063: Clarify memory size in 'neuro config show'
Currently, the output is
```
Name         #CPU    Memory   Preemptible   GPU
cpu-micro     0.1       100        ✖︎
cpu-small       1      4096        ✖︎
cpu-large       3     12288        ✖︎
gpu-small       7     30720        ✖︎        1 x nvidia-tesla-k80
gpu-large       7     30720        ✖︎        1 x nvidia-tesla-k80
```
Memory column is not obvious. I know that the size is in megabytes but `4096` looks obscure for a newbie. Humanized version, e.g. `4GB` looks better.

serhiy-storchaka:
Closes #1063.

### DIFF
--- a/neuromation/cli/formatters/config.py
+++ b/neuromation/cli/formatters/config.py
@@ -6,6 +6,7 @@ from tabulate import tabulate
 
 from neuromation.api.login import RunPreset
 from neuromation.cli.root import Root
+from neuromation.cli.utils import format_size
 
 
 class ConfigFormatter:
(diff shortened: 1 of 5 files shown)


shagren:
Why not use `humanize` which already used for file size humanization?

serhiy-storchaka:
It is use `humanize`.

shagren:
@serhiy-storchaka , sorry. Missed that

PR #1075, opened 2019-09-25
Parse public remote images by API, e.g. ubuntu:latest

asvetlov:


ayushkovskiy:
Basically, `ubuntu:latest` is local image and we should not get a reasonable result when parsing it via `remote_image()`. However, I think, we are getting to the point where merging two abstractions -- LocalImage and RemoteImage -- into a single `Image` is necessary:
```
@dataclass(frozen=True)
class Image:
    name: str
    tag: str
    owner: Optional[str] = None  # None for local image
    registry: Optional[str] = None  # None for local image
```
Thus, we can make these two methods private: `parse_as_neuro_image` and `parse_as_local_image`, and expose only `parse_remote` (named as just `parse`), which detects if the image `is_in_neuro_registry` then parse as a normal image `library/ubuntu:latest`, otherwise -- pase image as a neuro-image (using `image://` or `<registry-hostname>/`). But this is a big refactoring to be left for the future.

asvetlov:
> Basically, `ubuntu:latest` is local image

I disagree. When `ubuntu:latest` is used for `jobs.run()` in container specification it is definitely not a local name but an image hosted on public docker hub registry.

I've found that clear separation between local and remote image makes sense.

asvetlov:
push/pull operations have a semantic difference what is a local filesystem image and what is a remote one.
When you start a pod on kubernetis you always use *remote* image, doesn't matter what registry is used for it.

Issue #1093, opened 2019-10-01
Issues with copying a large tree


serhiy-storchaka:
I have encountered the following issues when copying a large tree:

1.
```
$ time neuro --show-traceback storage cp -r storage:fastai /tmp          
Copy storage://serhiystorchaka/fastai => file:///tmp/fastai
storage://serhiystorchaka/fastai/data/oxford-iiit-pet/annotations/xmls DONE
basset_hound_41.png [0.00%] 0B of 2.842K
._american_bulldog_152.png [0.00%] 0B of 240B
._British_Shorthair_52.png [0.00%] 0B of 240B
._Maine_Coon_71.png [0.00%] 0B of 240B
._english_cocker_spaniel_44.png [0.00%] 0B of 240B
._Russian_Blue_226.png [0.00%] 0B of 240B
._staffordshire_bull_terrier_110.png [0.00%] 0B of 240B
._Sphynx_29.png [0.00%] 0B of 240B
wheaten_terrier_9.png [0.00%] 0B of 2.305K
ERROR: Connection error (Response payload is not completed)
Traceback (most recent call last):
  File "/home/serhiy/neuromation/platform-client-python/neuromation/cli/main.py", line 288, in main
    cli.main(args=args, standalone_mode=False)
  File "/home/serhiy/neuromation/platform-client-python/venv/lib/python3.6/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/serhiy/neuromation/platform-client-python/venv/lib/python3.6/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/serhiy/neuromation/platform-client-python/venv/lib/python3.6/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/serhiy/neuromation/platform-client-python/venv/lib/python3.6/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/serhiy/neuromation/platform-client-python/venv/lib/python3.6/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/serhiy/neuromation/platform-client-python/venv/lib/python3.6/site-packages/click/decorators.py", line 27, in new_func
    return f(get_current_context().obj, *args, **kwargs)
[...]

PR #1107, opened 2019-10-08
Retry storage operations in case of ClientError.

serhiy-storchaka:
Can fix the part of #1093.

### DIFF
--- a/neuromation/api/storage.py
+++ b/neuromation/api/storage.py
@@ -1,8 +1,10 @@
 import asyncio
+import contextlib
 import datetime
 import enum
 import errno
 import fnmatch
+import logging
 import os
 import re
 import time
@@ -15,8 +17,10 @@ from typing import (
     AsyncIterator,
     Awaitable,
     Callable,
+    ContextManager,
     Dict,
     Iterable,
+    Iterator,
     List,
     Optional,
     Tuple,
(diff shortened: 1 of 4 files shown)


neuromation/api/storage.py
```diff
@@ -713,3 +743,22 @@ def leave(self, data: StorageProgressLeaveDir) -> None:
 
     def fail(self, data: StorageProgressFail) -> None:
         pass
+
+
+def retries(msg: str) -> Iterator[ContextManager[None]]:
+    sleeptime = 0.0
+    for r in range(ATTEMPTS)[::-1]:
```
asvetlov: Why is reversed iteration here?
I believe that the thing may be implemented simpler:
```
for r in range(ATTEMPTS):
    ...
    yield retry()
else:
    raise last_error
```

neuromation/api/storage.py
```diff
@@ -713,3 +743,22 @@ def leave(self, data: StorageProgressLeaveDir) -> None:
...
+        if r:
+
+            @contextlib.contextmanager
+            def retry() -> Iterator[None]:
```
asvetlov: Isn't the function should be *async*?
IIUC an exception raised during the work (not the first coroutine call but in the middle of the coroutine unwinding) is not handled.

It would be nice to have a test for `retries()` to check this.

neuromation/api/storage.py
serhiy-storchaka: Good idea!

neuromation/api/storage.py
serhiy-storchaka: You are right. There is `asyncio.sleep()` inside, so the context manager should be asynchronous.

PR #1121, opened 2019-10-21
Tracing

### LINKED ISSUE #467: Implement --verbose option that prints HTTP communcation
... `--verbose` or maybe `--debug`?

After refactoring we lost this functionality. Ideally, we could just set up the aiohttp client to print all the communication if `--verbose` was provided

asvetlov:
Fixes #467

### DIFF
--- a/neuromation/cli/root.py
+++ b/neuromation/cli/root.py
@@ -3,7 +3,7 @@ from dataclasses import dataclass
 from http.cookies import Morsel  # noqa
 from pathlib import Path
 from types import SimpleNamespace
-from typing import Dict, Optional, Tuple
+from typing import Dict, List, Optional, Tuple
 
 import aiohttp
 import click
@@ -85,6 +85,7 @@ class Root:
         return self._client
 
     async def init_client(self) -> None:
+        trace_configs: Optional[List[aiohttp.TraceConfig]]
         if self.trace:
             trace_configs = [self._create_trace_config()]
         else:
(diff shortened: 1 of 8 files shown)


neuromation/cli/root.py
```diff
@@ -79,11 +85,28 @@ def client(self) -> Client:
         return self._client
 
     async def init_client(self) -> None:
-        self._factory = Factory(path=self.config_path)
+        trace_configs: Optional[List[aiohttp.TraceConfig]]
+        if self.trace:
+            trace_configs = [self._create_trace_config()]
+        else:
+            trace_configs = None
```
serhiy-storchaka: Would not be simpler to set `trace_configs = []` here and drop `Optional` from annotations?

neuromation/cli/root.py
```diff
@@ -23,10 +28,11 @@ class Root:
     disable_pypi_version_check: bool
     network_timeout: float
     config_path: Path
+    trace: bool
```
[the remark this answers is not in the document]
asvetlov: The *real* code (not test) that creates `Root` instance is the only [here](https://github.com/neuromation/platform-client-python/blob/master/neuromation/cli/main.py#L208-L215)

I prefer to be explicit: if the code passes everything in the only existing call -- there is no need for defaults.

neuromation/cli/root.py
asvetlov: The only reason is that `ClientSession` has default `None` for `trace_configs`. The empty list works too, though.

PR #1130, opened 2019-10-23
Display Entrypoint in `neuro status` output

### LINKED ISSUE #924: No way to get entrypoint of job
Since #896 we have `--entrypoint` option. But `neuro ps` and `neuro status` will not show anything related to this new option. When user run:
```
neuro run --entrypoint sleep python 15m
```
and then share this job with teammate
Teammate will use `neuro status`. And him will see only  `Command: 15m`, which will be very strange for him.

shagren:
closes #924

### DIFF
--- a/CHANGELOG.D/924.feature
+++ b/CHANGELOG.D/924.feature
@@ -1,0 +1,1 @@
+Display `Entrypoint` field for job status command.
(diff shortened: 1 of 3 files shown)


asvetlov:
@shagren please add a change note and merge

PR #1135, opened 2019-10-28
port-forward --reconnect option

### LINKED ISSUE #1088: Add the option `neuro port-forward --forever`
use-case 1: the connection opened by `neuro port-forward` has failed,

use-case 2: the user closes the laptop lid and returns back to work in 5 minutes (the job remains running).

In both cases, the user has to find the terminal where the command was executed and re-run it, otherwise, remote debugging via PyCharm won't work. Instead, we could re-run `neuro port-forward` automatically.

shagren:
closes #1088

### DIFF
--- a/neuromation/api/jobs.py
+++ b/neuromation/api/jobs.py
@@ -645,7 +645,7 @@ class _AwaitDisconnection:
 
     async def _wait(self) -> None:
         # the tranpoline is needed to shift the return type from int to None
-        self._proc.wait()
+        await self._proc.wait()
 
     def __await__(self) -> Generator[Any, None, None]:
         return self._wait().__await__()
--- a/neuromation/api/jobs.py
+++ b/neuromation/api/jobs.py
@@ -371,6 +371,10 @@ class Jobs(metaclass=NoPublicConstructor):
             f"ProxyCommand={proxy_command_str}",
             "-o",
             "ExitOnForwardFailure=yes",
+            "-o",
+            "ServerAliveInterval=5",
+            "-o",
+            "ServerAliveCountMax=2",
         ]
         if no_key_check:  # pragma: no branch
             command += [
@@ -384,8 +388,7 @@ class Jobs(metaclass=NoPublicConstructor):
         try:
             result = await proc.wait()
             if result != 0:
-                raise ValueError(f"error code {result}")
-            return
+                raise SystemExit(result)
         finally:
             await _kill_proc_tree(proc.pid, timeout=10)
             # add a sleep to get process watcher a chance to execute all callbacks
--- a/neuromation/api/jobs.py
+++ b/neuromation/api/jobs.py
@@ -3,6 +3,7 @@ import enum
 import json
 import shlex
 import signal
+from contextlib import suppress
 from dataclasses import dataclass, field
 from datetime import datetime
 from typing import Any, AsyncIterator, Dict, Iterable, List, Mapping, Optional, Sequence
@@ -39,7 +40,7 @@ from .parsing_utils import (
     _ImageNameParser,
     _is_in_neuro_registry,
 )
-from .utils import NoPublicConstructor
+from .utils import NoPublicConstructor, asynccontextmanager
 
 
 INVALID_IMAGE_NAME = "INVALID-IMAGE-NAME"
@@ -311,25 +312,28 @@ class Jobs(metaclass=NoPublicConstructor):
             # add a sleep to get process watcher a chance to execute all callbacks
             await asyncio.sleep(0.1)
 
-    def port_forward(
+    @asynccontextmanager
+    async def port_forward(
         self, id: str, local_port: int, job_port: int, *, no_key_check: bool = False
-    ) -> "asyncio.Task[int]":
+    ) -> AsyncIterator["asyncio.Task[None]"]:
         loop = asyncio.get_event_loop()
         task = loop.create_task(
             self._port_forward(id, local_port, job_port, no_key_check=no_key_check)
         )
-        return task
+        yield task
+        task.cancel()
+        with suppress(asyncio.CancelledError):
+            await task
 
     async def _port_forward(
         self, id: str, local_port: int, job_port: int, *, no_key_check: bool = False
-    ) -> int:
+    ) -> None:
         try:
             job_status = await self.status(id)
         except IllegalArgumentError as e:
             raise ValueError(f"Job not found. Job Id = {id}") from e
-        if job_status.status != JobStatus.RUNNING:
+        if job_status.status != "running":
             raise ValueError(f"Job is not running. Job Id = {id}")
-
         payload = json.dumps(
             {
                 "method": "job_port_forward",
@@ -385,7 +389,8 @@ class Jobs(metaclass=NoPublicConstructor):
         command += [f"{server_url.user}@{server_url.host}"]
         proc = await asyncio.create_subprocess_exec(*command)
         try:
-            return await proc.wait()
+            result = await proc.wait()
+            raise ConnectionError(f"ssh exited with code {result}")
         finally:
             await _kill_proc_tree(proc.pid, timeout=10)
             # add a sleep to get process watcher a chance to execute all callbacks
--- a/neuromation/api/jobs.py
+++ b/neuromation/api/jobs.py
@@ -3,7 +3,7 @@ import enum
 import json
 import shlex
 import signal
-from contextlib import suppress
+from asyncio import Task
 from dataclasses import dataclass, field
 from datetime import datetime
 from typing import Any, AsyncIterator, Dict, Iterable, List, Mapping, Optional, Sequence
@@ -312,18 +312,14 @@ class Jobs(metaclass=NoPublicConstructor):
             # add a sleep to get process watcher a chance to execute all callbacks
             await asyncio.sleep(0.1)
 
-    @asynccontextmanager
-    async def port_forward(
+    def port_forward(
         self, id: str, local_port: int, job_port: int, *, no_key_check: bool = False
-    ) -> AsyncIterator[None]:
+    ) -> Task:
         loop = asyncio.get_event_loop()
         task = loop.create_task(
             self._port_forward(id, local_port, job_port, no_key_check=no_key_check)
         )
-        yield
-        task.cancel()
-        with suppress(asyncio.CancelledError):
-            await task
+        return task
 
     async def _port_forward(
         self, id: str, local_port: int, job_port: int, *, no_key_check: bool = False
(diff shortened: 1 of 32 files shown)


neuromation/cli/job.py
```diff
@@ -365,14 +367,19 @@ def job() -> None:
 @command()
 @click.argument("job")
 @click.argument("local_remote_port", type=LOCAL_REMOTE_PORT, nargs=-1, required=True)
+@click.option("--reconnect/--no-reconnect", default=True, is_flag=True, help="TODO")
```
serhiy-storchaka: I think `is_flag=True` is redundant if we use "--foo/--no-foo".

neuromation/cli/job.py
```diff
@@ -393,25 +400,69 @@ def job() -> None:
...
-                )
-            )
 
-        click.echo("Press ^C to stop forwarding")
-        try:
-            while True:
-                await asyncio.sleep(1)
-        except KeyboardInterrupt:
-            pass
+    from asyncio import Task
```
serhiy-storchaka: Import it at top level.

neuromation/cli/job.py
```diff
@@ -393,25 +400,69 @@ def job() -> None:
...
+
+    _tasks: List[Task] = []
```
serhiy-storchaka: Since all these names are local variable, there is need to underscore them. Underscores reduce readability.

neuromation/cli/job.py
```diff
@@ -393,25 +400,69 @@ def job() -> None:
...
+
+        if not _stopped and reconnect:
+            if task.exception():
+                if type(task.exception()) is not ValueError:
+                    _forward(local_remote_port)
+                else:
+                    _stop()
+
+    def _forward(local_remote_port: Tuple[int, int]):
+        nonlocal _tasks
```
serhiy-storchaka: There is no assignment to `_tasks`. `nonlocal` is not needed.

neuromation/cli/job.py
```diff
@@ -393,25 +400,69 @@ def job() -> None:
...
+        print("Done: ", local_remote_port, task)
+
+        if not _stopped and reconnect:
+            if task.exception():
+                if type(task.exception()) is not ValueError:
+                    _forward(local_remote_port)
+                else:
+                    _stop()
+
+    def _forward(local_remote_port: Tuple[int, int]):
```
serhiy-storchaka: Why not just two parameters local_port and job_port? `local_remote_port[0]` below looks ugly.

neuromation/cli/job.py
```diff
@@ -393,25 +400,69 @@ def job() -> None:
...
-                await asyncio.sleep(1)
-        except KeyboardInterrupt:
-            pass
+    from asyncio import Task
+
+    _tasks: List[Task] = []
+    _stopped: bool = False
+
+    def _done(local_remote_port: Tuple[int, int], task: Task):
+        nonlocal reconnect
```
serhiy-storchaka: None of these variables need `nonlocal`.

neuromation/cli/job.py
```diff
@@ -393,25 +400,69 @@ def job() -> None:
...
-            while True:
-                await asyncio.sleep(1)
-        except KeyboardInterrupt:
-            pass
+    from asyncio import Task
+
+    _tasks: List[Task] = []
+    _stopped: bool = False
+
+    def _done(local_remote_port: Tuple[int, int], task: Task):
```
serhiy-storchaka: `done()` could be defined as local function inside `forward()`, so you would not need to pass `local_remote_port` and use `partial()`, because it is available as nonlocal.

neuromation/cli/job.py
```diff
@@ -393,25 +400,69 @@ def job() -> None:
...
+        nonlocal reconnect
+        nonlocal _stopped
+        nonlocal _tasks
+
+        print("Done: ", local_remote_port, task)
+
+        if not _stopped and reconnect:
+            if task.exception():
+                if type(task.exception()) is not ValueError:
```
serhiy-storchaka: Why not `isinstance()`?

neuromation/cli/job.py
```diff
@@ -393,25 +400,69 @@ def job() -> None:
...
+        print("Forward: ", local_remote_port)
+
+        task = root.client.jobs.port_forward(
+            job_id,
+            local_remote_port[0],
+            local_remote_port[1],
+            no_key_check=no_key_check,
+        )
+        _tasks.append(task)
+        task.add_done_callback(partial(_done, local_remote_port))
```
serhiy-storchaka: I would add `if reconnect:` before `task.add_done_callback()` and remove checking `reconnect` from `done()`.

neuromation/cli/job.py
```diff
@@ -393,25 +400,69 @@ def job() -> None:
...
+    click.echo("Press ^C to stop forwarding")
+    try:
+        while not _stopped:
+            await asyncio.sleep(1)
+
+    except KeyboardInterrupt:
+        _stopped = True
+        pass
+
+    for task in _tasks:
```
serhiy-storchaka: Can `gather()` or `wait()` be used here?

neuromation/cli/job.py
```diff
@@ -393,25 +400,69 @@ def job() -> None:
...
+        with suppress(asyncio.CancelledError):
+            await task
+        _tasks.pop(_tasks.index(task))
```
serhiy-storchaka: Why this? Iterating a modified list can be surprised. In particular you will skip tasks at odd positions.

neuromation/cli/job.py
shagren: This PR not ready :)

neuromation/cli/job.py
shagren: No, i have checked now :(

neuromation/cli/job.py
shagren: not actual more

neuromation/cli/job.py
shagren: i have tried but something went wrong :(

neuromation/cli/job.py
```diff
@@ -366,14 +366,19 @@ def job() -> None:
 @command()
 @click.argument("job")
 @click.argument("local_remote_port", type=LOCAL_REMOTE_PORT, nargs=-1, required=True)
+@click.option("--reconnect/--no-reconnect", default=True, is_flag=True, help="TODO")
```
shagren: Need some ideas here

neuromation/api/jobs.py
```diff
@@ -312,36 +311,41 @@ def __init__(self, core: _Core, config: _Config) -> None:
...
                 "params": {"job": id, "port": job_port},
             }
         )
-        proxy_command = ["ssh"]
+        proxy_command = [
+            "ssh",
+            "-o",
+            "ExitOnForwardFailure=yes",
+            "-o",
+            "ServerAliveInterval=10",
```
shagren: In fact this options didn't help so much :(

neuromation/api/jobs.py
```diff
@@ -312,36 +311,41 @@ def __init__(self, core: _Core, config: _Config) -> None:
             # add a sleep to get process watcher a chance to execute all callbacks
             await asyncio.sleep(0.1)
 
-    @asynccontextmanager
-    async def port_forward(
+    def port_forward(
         self, id: str, local_port: int, job_port: int, *, no_key_check: bool = False
-    ) -> AsyncIterator[None]:
+    ) -> "asyncio.Task[int]":
```
shagren: Will return Task. User must cancel task self when need.

neuromation/cli/job.py
```diff
@@ -394,25 +399,59 @@ def job() -> None:
...
+        except ValueError:  # Job not found
+            break
+
+        if not reconnect:
+            return
+
+        for num, task_definition in enumerate(tasks):
+            task = task_definition.task
+            if task in done:
+                if isinstance(task.exception(), ValueError):
```
shagren: Only ValueError (task not found/ task not running) will stop execution

neuromation/cli/utils.py
```diff
@@ -458,6 +460,8 @@ def volume_to_verbose_str(volume: Volume) -> str:
     try:
         jobs = await client.jobs.list(name=id_or_name, owners={owner})
     except Exception as e:
+        if raise_error:
+            raise e
```
ayushkovskiy: I would remove this change: the use-case of `resolve_job()` for `port-forward` is the same as for all other commands, thus there should be no difference in its logic. Moreover, if we re-raise the exception `e` caught from `client.jobs.list()`, it will be confusingly printed directly to the console.

shagren: Removed

neuromation/cli/job.py
```diff
@@ -1059,3 +1074,27 @@ def calc_statuses(status: Sequence[str], all: bool) -> Set[JobStatus]:
...
+    root: Root, job_id: str, local_port: int, job_port: int, no_key_check: bool
+) -> AsyncIterator[None]:
+    while True:
+        try:
+            async with root.client.jobs.port_forward(
+                job_id, local_port, job_port, no_key_check=no_key_check
+            ):
+                yield
+        except (ConnectionError, ClientConnectionError):
+            click.echo(f"Reconnect localhost:{local_port} => {job_id}:{job_port}")
```
ayushkovskiy: feature idea: let's print time when the reconnect happened? otherwise the user will have a long list of identical messages.

neuromation/api/jobs.py
```diff
@@ -341,7 +341,15 @@ def __init__(self, core: _Core, config: _Config) -> None:
                 "params": {"job": id, "port": job_port},
             }
         )
-        proxy_command = ["ssh"]
+        proxy_command = [
```
ayushkovskiy: if these changes didn't help much (https://github.com/neuromation/platform-client-python/pull/1135/files#r345802769) , do we need them?

asvetlov: They help a little maybe?
A little is better than nothing.

neuromation/api/jobs.py
```diff
@@ -623,3 +635,17 @@ def inner() -> None:
...
+
+class _AwaitDisconnection:
+    __slots__ = ("_proc",)
+
+    def __init__(self, proc: Process) -> None:
+        self._proc = proc
+
+    async def _wait(self) -> None:
+        # the tranpoline is needed to shift the return type from int to None
+        self._proc.wait()
```
shagren:
```suggestion
        await self._proc.wait()
```
because warning here:
>/home/shagren/PycharmProjects/platform-client-python/neuromation/api/jobs.py:648: RuntimeWarning: coroutine 'Process.wait' was never awaited
  self._proc.wait()
RuntimeWarning: Enable tracemalloc to get the object allocation traceback

asvetlov: Thanks!

yartem:
Should be done in https://github.com/neuromation/platform-client-python/pull/1555

PR #1143, opened 2019-10-30
Distributed tracing

### LINKED ISSUE #706: Provide a token for distributed tracing
Every HTTP operation for API should send `X-NEURO-TRACE-ID` header.
The value is a UUID string unique for every neuro CLI run.

It allows using distributing tracing on server side, e.g. Zipkin/aiozipkin.
Merging all requests from the single CLI run is an important facility.

*UPD*
Zipkin (and open tracing) supports `X-B3-TraceId`, `X-B3-SpanId`, `X-B3-ParentSpanId`, `X-B3-Flags`, `X-B3-Sampled` HTTP headers. Need to investigate more and use standard names.

asvetlov:
Fixes #706

Send zipkin-compatible tracing 'b3' header to servers.
span-id is regenerated for every HTTP request, where is no choice here.
sampling-state is not sent, it is a server duty to decide what sampling rate is appropriate.
trace-id is regenerated for every request if not specified explicitly by api `Factory` construction.

CLI uses unique trace-id during the whole CLI run; it allows to merge all subsequent HTTTP calls for batch operations like recursive file uploading. 
Not sure how to achieve this for programmatic API usage; introducing additional layer just for grouping is tedious:
```
async with get() as client:
   async with client.new_span() as span():
       await span.jobs.run(....)
```

Do we need it at all?
Ideas?

### DIFF
--- a/neuromation/api/config_factory.py
+++ b/neuromation/api/config_factory.py
@@ -26,6 +26,7 @@ from .login import (
     get_server_config,
     refresh_token,
 )
+from .tracing import _make_trace_config
 from .utils import _ContextManager
 
 
@@ -63,11 +64,15 @@ class Factory:
         self,
         path: Optional[Path] = None,
         trace_configs: Optional[List[aiohttp.TraceConfig]] = None,
+        trace_id: Optional[str] = None,
     ) -> None:
         if path is None:
             path = Path(os.environ.get(CONFIG_ENV_NAME, DEFAULT_CONFIG_PATH))
         self._path = path.expanduser()
-        self._trace_configs = trace_configs
+        self._trace_configs = [_make_trace_config()]
+        if trace_configs:
+            self._trace_configs += list(trace_configs)
+        self._trace_id = trace_id
 
     async def get(self, *, timeout: aiohttp.ClientTimeout = DEFAULT_TIMEOUT) -> Client:
         saved_config = config = self._read()
@@ -98,7 +103,7 @@ class Factory:
             await session.close()
             raise
         else:
-            return Client._create(session, config)
+            return Client._create(session, config, self._trace_id)
 
     async def login(
         self,
--- a/neuromation/api/config_factory.py
+++ b/neuromation/api/config_factory.py
@@ -71,7 +71,7 @@ class Factory:
         self._path = path.expanduser()
         self._trace_configs = [_make_trace_config()]
         if trace_configs:
-            self._trace_configs += list(trace_configs)
+            self._trace_configs += trace_configs
         self._trace_id = trace_id
 
     async def get(self, *, timeout: aiohttp.ClientTimeout = DEFAULT_TIMEOUT) -> Client:
--- a/neuromation/api/core.py
+++ b/neuromation/api/core.py
@@ -2,6 +2,7 @@ import errno
 import json as jsonmodule
 import logging
 from http.cookies import Morsel  # noqa
+from types import SimpleNamespace
 from typing import Any, AsyncIterator, Dict, Mapping, Optional
 
 import aiohttp
@@ -9,6 +10,7 @@ from aiohttp import WSMessage
 from multidict import CIMultiDict
 from yarl import URL
 
+from .tracing import gen_trace_id
 from .utils import asynccontextmanager
 
 
@@ -57,10 +59,12 @@ class _Core:
         base_url: URL,
         token: str,
         cookie: Optional["Morsel[str]"],
+        trace_id: Optional[str],
     ) -> None:
         self._session = session
         self._base_url = base_url
         self._token = token
+        self._trace_id = trace_id
         self._headers = self._auth_headers()
         if cookie is not None:
             self._session.cookie_jar.update_cookies(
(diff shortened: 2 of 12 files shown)


neuromation/api/tracing.py
```diff
@@ -0,0 +1,42 @@
...
+
+    The id is used for distributed tracing.
+    """
+    return str(binascii.hexlify(os.urandom(8)).decode("utf-8"))
+
+
+def _update_headers(headers: "CIMultiDict[str]", trace_id: str, span_id: str) -> None:
+    """Creates dict with zipkin single header format.
+    """
+    # b3={TraceId}-{SpanId}-{SamplingState}-{ParentSpanId}
```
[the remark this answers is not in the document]
asvetlov: It works only if this parent span is stored in DB.
Since we don't save to zipkin from a client the span doesn't exist.
Recursive upload is still grouped together because everything shares the same trace id.
I believe it should be enough.
Server code *will* use parent span id feature, e.g. for requesting auth microservice.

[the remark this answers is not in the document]
dalazx: interesting. I might be missing smith. I was thinking about the following use case:
we want to download a directory. first, we get a file listing, then we spawn tasks and pull the data based on the listing. is this also a legit case for leveraging parent spans?

neuromation/api/tracing.py
```diff
@@ -0,0 +1,42 @@
...
+    trace_id = getattr(trace_ctx, "trace_id", None)
+    if trace_id is None:
+        return
+    span_id = gen_trace_id()
+    _update_headers(params.headers, trace_id, span_id)
+
+
+def _make_trace_config() -> aiohttp.TraceConfig:
+    trace_config = aiohttp.TraceConfig()
+    trace_config.on_request_start.append(_on_request_start)  # type: ignore
```
serhiy-storchaka: Why ignore?

neuromation/api/config_factory.py
```diff
@@ -63,11 +64,15 @@ def __init__(
...
         trace_configs: Optional[List[aiohttp.TraceConfig]] = None,
+        trace_id: Optional[str] = None,
     ) -> None:
         if path is None:
             path = Path(os.environ.get(CONFIG_ENV_NAME, DEFAULT_CONFIG_PATH))
         self._path = path.expanduser()
-        self._trace_configs = trace_configs
+        self._trace_configs = [_make_trace_config()]
+        if trace_configs:
+            self._trace_configs += list(trace_configs)
```
serhiy-storchaka: Explicit converting to list is not needed.

neuromation/api/registry.py
```diff
@@ -14,10 +14,15 @@ class _Registry(_Core):
     """
 
     def __init__(
-        self, session: aiohttp.ClientSession, base_url: URL, token: str, username: str
+        self,
+        session: aiohttp.ClientSession,
+        base_url: URL,
+        token: str,
+        trace_id: Optional[str],
```
serhiy-storchaka: Why `trace_id` was inserted between `token` and `username`?

neuromation/api/config_factory.py
asvetlov: thanks

neuromation/api/registry.py
asvetlov: The constructor is *internal*, the order doesn't matter.
But I've chosen to use session-base_url-token-trace_id sequence because base class `_Core` has this signature

neuromation/api/tracing.py
```diff
@@ -0,0 +1,42 @@
...
+import binascii
+import os
+import types
+
+import aiohttp
+from multidict import CIMultiDict
+
+
+def gen_trace_id() -> str:
+    """Returns a 64 bit UTF-8 encoded string.
```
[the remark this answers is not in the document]
asvetlov: The docstring is borrowed from aiozipkin which in turn is borrowed from the synchronous zipkin client.
Your suggestion is much better, sure.

neuromation/api/tracing.py
asvetlov: I suspect it is mypy bug, very close to your recent report:
```
neuromation/api/tracing.py:40: error: Argument 1 to "append" of "MutableSequence" has incompatible type "Callable[[ClientSession, SimpleNamespace, TraceRequestStartParams], Coroutine[Any, Any, None]]"; expected "Callable[[ClientSession, SimpleNamespace, <union: 15 items>], Awaitable[None]]"
```
Coroutine is not resolved as Awaitable or vice versa.

neuromation/api/core.py
```diff
@@ -112,6 +116,11 @@ def _auth_headers(self) -> Dict[str, str]:
         else:
             real_headers = CIMultiDict()
         real_headers.update(self._headers)
+        trace_request_ctx = SimpleNamespace()
+        trace_id = self._trace_id
+        if trace_id is None:
+            trace_id = gen_trace_id()
+        trace_request_ctx.trace_id = trace_id
```
shagren: Why you generate unique-per-request trace_id here?

asvetlov: What do you propose?

PR #1152, opened 2019-11-05
Pager for output lists

asvetlov:
The output can be potentially long, the pager is auto-applied in this case.


asvetlov:
All comments are fixed.
@serhiy-storchaka thanks for the review!

Issue #1141, opened 2019-10-30
Introduce `neuro config quota show`


dalazx:
We need to be able to communicate to our users about their quota situation.

`neuro config quota show` should render the current jobs run-time consumed and quota for the current user's cluster.

The information should be retrieved using https://github.com/neuromation/platform-api/issues/901.

Optionally, the command should output a suggestion for topping up the quota.

cc @asvetlov @mariyadavydova

asvetlov:
It can live in `config` group for now.
When we implement organizations we can reconsider the command position, moving with the deprecation of previous notation is not a big deal.

ayushkovskiy:
Guys are we OK with this output format?
```
GPU left: 167h 23m 32s (quota: 200h 00m 00s)
CPU left:  00h 00m 32s (quota: 1000h 00m 00s)
```

mariyadavydova:
@ayushkovskiy 👍

ayushkovskiy:
Huh, we don't provide seconds for quota.. So the formatting will look like:
```
GPU left: 167h 23m (quota: 200h 00m)
CPU left:  00h 00m (quota: 1000h 00m)
```
@mariyadavydova HDYT wouldn't it look weird without seconds?

mariyadavydova:
@ayushkovskiy Well, that's OK. I don't think that seconds really matter here, when we speak about 100h quotas.

ayushkovskiy:
How do we format lack of quota? `(quota: NaN)` or `(quota: None)` or `(quota: ∞)`?

mariyadavydova:
The last one is the most accurate, I guess.

ayushkovskiy:
the sign looks too tiny

mariyadavydova:
We may simply say `(quota: infinite)`. I don't think that this question is very important ATM, though, because those who have no quota (i.e. non-free-tier folks) will hardly ever execute this command.

asvetlov:
+1 for 'infinity'. Not all terminals support Unicode symbols.

ayushkovskiy:
if `quota == infinity`, `left` is also `infinity`. Let's stick to this formatting:
```
GPU spent: 167h 23m (quota: 170h 23m, left: 03h 00m)
CPU spent:  00h 00m  (quota: infinity)
```

PR #1156, opened 2019-11-08
Introduce `neuro config quota show`

ayushkovskiy:
Closes https://github.com/neuromation/platform-client-python/issues/1141

### DIFF
--- a/neuromation/api/client.py
+++ b/neuromation/api/client.py
@@ -6,6 +6,8 @@ from typing import Mapping, Optional, Type
 
 import aiohttp
 
+from neuromation.api.quota import Quota
+
 from .config import _Config
 from .core import _Core
 from .images import Images
@@ -46,6 +48,7 @@ class Client(metaclass=NoPublicConstructor):
         self._storage = Storage._create(self._core, self._config)
         self._users = Users._create(self._core)
         self._parser = Parser._create(self._config, self.username)
+        self._quota = Quota._create(self._core, self._config)
         self._images: Optional[Images] = None
 
     async def close(self) -> None:
@@ -97,6 +100,10 @@ class Client(metaclass=NoPublicConstructor):
     @property
     def parse(self) -> Parser:
         return self._parser
+
+    @property
+    def quota(self) -> Quota:
+        return self._quota
 
     def _get_session_cookie(self) -> Optional["Morsel[str]"]:
         for cookie in self._core._session.cookie_jar:
--- a/neuromation/api/client.py
+++ b/neuromation/api/client.py
@@ -6,7 +6,7 @@ from typing import Mapping, Optional, Type
 
 import aiohttp
 
-from neuromation.api.quota import Quota
+from neuromation.api.quota import _Quota
 
 from .config import _Config
 from .core import _Core
@@ -48,7 +48,7 @@ class Client(metaclass=NoPublicConstructor):
         self._storage = Storage._create(self._core, self._config)
         self._users = Users._create(self._core)
         self._parser = Parser._create(self._config, self.username)
-        self._quota = Quota._create(self._core, self._config)
+        self._quota = _Quota._create(self._core, self._config)
         self._images: Optional[Images] = None
 
     async def close(self) -> None:
@@ -100,10 +100,6 @@ class Client(metaclass=NoPublicConstructor):
     @property
     def parse(self) -> Parser:
         return self._parser
-
-    @property
-    def quota(self) -> Quota:
-        return self._quota
 
     def _get_session_cookie(self) -> Optional["Morsel[str]"]:
         for cookie in self._core._session.cookie_jar:
--- a/neuromation/api/quota.py
+++ b/neuromation/api/quota.py
@@ -1,0 +1,41 @@
+from dataclasses import dataclass
+from typing import Any, Dict, Optional
+
+from yarl import URL
+
+from neuromation.api.config import _Config
+from neuromation.api.core import _Core
+from neuromation.api.utils import NoPublicConstructor
+
+
+@dataclass(frozen=True)
+class QuotaInfo:
+    name: str
+    spent_gpu_minutes: int
+    spent_non_gpu_minutes: int
+    quota_gpu_minutes: Optional[int]
+    quota_non_gpu_minutes: Optional[int]
+
+
+class Quota(metaclass=NoPublicConstructor):
+    def __init__(self, core: _Core, config: _Config) -> None:
+        self._core = core
+        self._config = config
+
+    async def get(self) -> QuotaInfo:
+        url = URL(f"stats/users/{self._config.auth_token.username}")
+        async with self._core.request("GET", url) as resp:
+            res = await resp.json()
+            return _quota_info_from_api(res)
+
+
+def _quota_info_from_api(payload: Dict[str, Any]) -> QuotaInfo:
+    total_gpu_str = payload["quota"].get("total_gpu_run_time_minutes")
+    total_non_gpu_str = payload["quota"].get("total_non_gpu_run_time_minutes")
+    return QuotaInfo(
+        name=payload["name"],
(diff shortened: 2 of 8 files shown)


ayushkovskiy:
`GET /stats/users/{name}` is currently deployed only on `dev: default` cluster only

neuromation/api/quota.py
```diff
@@ -0,0 +1,59 @@
...
+from yarl import URL
+
+from neuromation.api.config import _Config
+from neuromation.api.core import _Core
+from neuromation.api.utils import NoPublicConstructor
+
+
+@dataclass(frozen=True)
+class QuotaDetails:
+    spent_minutes: int
```
asvetlov: Don't use minutes in API but float seconds.
float("inf") is for infinity, please use this constant instead of `None`

neuromation/api/quota.py
```diff
@@ -0,0 +1,59 @@
...
+    gpu_details: QuotaDetails
+    cpu_details: QuotaDetails
+
+
+class Quota(metaclass=NoPublicConstructor):
+    def __init__(self, core: _Core, config: _Config) -> None:
+        self._core = core
+        self._config = config
+
+    async def get(self) -> QuotaInfo:
```
asvetlov: How to get quota for another user?
I can do it if have appropriate permissions, right?

neuromation/cli/config.py
```diff
@@ -28,6 +29,11 @@ def config() -> None:
     """Client configuration."""
 
 
+@group()
+def quota() -> None:
```
asvetlov: Don't use nested group, `neuro config show-quota` is just fine.

ayushkovskiy: AFAIK, the team decided to use `neuro config quota show` in order to extend `quota` group with additional funcitonality: `neuro config quota add`, etc. Perhaps, the good solution would be to make the group `quota` non-nested: `neuro quota show`

neuromation/api/quota.py
ayushkovskiy: do we really need a funcitonality to see other user's quota? cc @mariyadavydova

neuromation/cli/config.py
asvetlov: We decided to put quota management into the config group, that's it.
Please don't add redundant nesting.

neuromation/api/quota.py
ayushkovskiy: > Don't use minutes in API but float seconds.

for this, we will need to change the server-side API. Note, currently, we get the time in minutes (see https://github.com/neuromation/platform-api/issues/901).

Should we change the API?

neuromation/api/quota.py
asvetlov: We obviously support it by REST API. Otherwise please modify the server as well.

neuromation/api/quota.py
asvetlov: Not necessary, you can multiply by 60 on the client.
In python there are two time representations: float and datetime.
It is confusing enough right now, adding yet another notation makes everything even worse and less usable: people will always convert minutes into somethind more convinient.

serhiy-storchaka: Why do we need `QuotaDetails`? Why not use just `float` or `Optional[timedelta]`?

neuromation/cli/config.py
serhiy-storchaka: I think the common user cannot change quota.

asvetlov: Also please keep in mind: later, after organizations implementing, we may find that quota better fits somewhere in the organization management structures. Now we don't add top-level `neuro quota` but put a simple command into the config group for this reason.

neuromation/api/quota.py
mariyadavydova: Ideally, yes, we need a functionality to see other user's quota (for example, I would like to be able to see the quotas left for our free tier users). However, `config` seems to be a bad place for such functionality, as other `config` commands refer to your personal information (or your server's). Let's file another issue and think where we should put the quota-related functionality (this refers to other discussions in this PR too). This particular issue should solve the immediate problem - "How do I, free tier user, get to know how much quota I have?".

mariyadavydova: > We obviously support it by REST API. Otherwise please modify the server as well.

Please, don't :) I use this functionality to bump quotas via REST API.

neuromation/api/quota.py
ayushkovskiy: > Not necessary, you can multiply by 60 on the client.

OK, I will change `..._minutes: int` to `..._seconds: float` everywhere. But for me personally it'd be confusing if I see `GPU spent: 00h 01m 00s`, then I run a job for 30 seconds and I still see: `GPU spent: 00h 01m 00s`. I think, if we need to show time in seconds, the API protocol should be changed.

ayushkovskiy: > Why do we need QuotaDetails? Why not use just float or Optional[timedelta]?

In the first version of my code, there were just fields `gpu_spent_minutes`, `cpu_limit_minutes`, etc. But since the lines for `CPU` and `GPU` should use the same formatting method `_format_quota_details`, I found it useful to introduce an additional abstraction `QuotaDetails`. Though, if it's an eyesore for you, I'm OK to get rid of it. Should I? cc @serhiy-storchaka

neuromation/api/quota.py
asvetlov: @mariyadavydova here is a public API class, not `neuro config` CLI command group.
API is more generic (and potentially harder to change publishing).

neuromation/api/quota.py
asvetlov: API uses seconds (omit the suffix from field names please).
CLI can stay with minutes, it makes sense.

Like we return file size as bytes in the storage API but show it as gigabytes for large files in CLI.

neuromation/cli/config.py
ayushkovskiy: OK

neuromation/api/quota.py
asvetlov: Flat structure with short field names is better IMHO

neuromation/api/quota.py
mariyadavydova: Oh, I'm sorry. Then I think that we should update this class accordingly and keep it as is for a while, as this quota functionality is emerging now and we do not have a clear understanding of how it should be organized.

asvetlov: `async def get(self, user: Optional[str]=None)`, the default `None` is for self.

neuromation/api/quota.py
ayushkovskiy: > API uses seconds (omit the suffix from field names please). CLI can stay with minutes, it makes sense.

Fixed.

> Why do we need QuotaDetails? Why not use just float or Optional[timedelta]?

fixed to use a flat structure.

neuromation/api/quota.py
ayushkovskiy: already done

ayushkovskiy:
All comments considered, everyone thanks for your review!

ayushkovskiy:
Currently adding the documentation on API

ayushkovskiy:
1. Added changelog message
2. Hide API for quota in order to not expose API that will likely change soon. Now, dataclass `_QuotaInfo` is protected, and `class Client` has a temporary protected method `_get_quota_info()`

neuromation/api/client.py
```diff
@@ -103,3 +106,27 @@ def _get_session_cookie(self) -> Optional["Morsel[str]"]:
             if cookie.key == "NEURO_SESSION":
                 return cookie
         return None
+
+    async def _get_quota_info(self, user: Optional[str] = None) -> _QuotaInfo:
```
asvetlov: Please keep `client._quota` private API.
There is no need to mess `client.py` with quota-specific things.

ayushkovskiy: OK

PR #1158, opened 2019-11-12
Remove user's token when `--trace` is enabled

ayushkovskiy:
Sorry for preparing the PR without an issue: I decided that this patch is small enough to avoid the bureaucracy.

Here's a small security issue: we need to hide user's sensitive data when printing out HTTP communication when `--trace` is enabled.

Thus, instead of 
```
$ neuro --trace run ubuntu
> GET /api/v1/config HTTP/1.1
> Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c
> 
...
```
the user would get 
```
$ neuro --trace run ubuntu
> GET /api/v1/config HTTP/1.1
> Authorization: Bearer <token>
> 
...
```

Thus, when the user is asked to re-run a command with `--trace` and share it in a bug tracker, they can be sure their sensitive data won't be revealed.

Disclaimer: all tokens used in description are fake.

### DIFF
--- a/neuromation/cli/main.py
+++ b/neuromation/cli/main.py
@@ -175,18 +175,18 @@ def print_options(
 )
 @click.option(
     "--trace",
+    default=None,
     is_flag=True,
     help="Trace sent HTTP requests and received replies to stderr.",
 )
 @click.option(
     "--hide-token/--no-hide-token",
     is_flag=True,
-    default=True,
-    show_default=True,
+    default=None,
     help=(
         "Prevent user's token sent in HTTP headers from being "
         "printed out to stderr during HTTP tracing. Can be used only "
-        "together with option '--trace'."
+        "together with option '--trace'  [default: True]"
     ),
 )
 @click.pass_context
--- a/neuromation/cli/main.py
+++ b/neuromation/cli/main.py
@@ -181,7 +181,8 @@ def print_options(
 @click.option(
     "--hide-token/--no-hide-token",
     is_flag=True,
-    default=None,
+    default=True,
+    show_default=True,
     help=(
         "Prevent user's token sent in HTTP headers from being "
         "printed out to stderr during HTTP tracing. Can be used only "
--- a/neuromation/cli/main.py
+++ b/neuromation/cli/main.py
@@ -185,7 +185,7 @@ def print_options(
     help=(
         "Prevent user's token sent in HTTP headers from being "
         "printed out to stderr during HTTP tracing. Can be used only "
-        "together with option '--trace'  [default: True]"
+        "together with option '--trace'. On by default."
     ),
 )
 @click.pass_context
--- a/neuromation/cli/main.py
+++ b/neuromation/cli/main.py
@@ -178,6 +178,16 @@ def print_options(
     is_flag=True,
     help="Trace sent HTTP requests and received replies to stderr.",
 )
+@click.option(
+    "--hide-token/--no-hide-token",
+    is_flag=True,
+    default=None,
+    help=(
+        "Prevent user's token sent in HTTP headers from being "
+        "printed out to stderr during HTTP tracing. Can be used only "
+        "together with option '--trace'."
+    ),
+)
 @click.pass_context
 def cli(
     ctx: click.Context,
@@ -189,6 +199,7 @@ def cli(
     disable_pypi_version_check: bool,
     network_timeout: float,
     trace: bool,
+    hide_token: Optional[bool],
 ) -> None:
     #   ▇ ◣
     #   ▇ ◥ ◣
@@ -211,6 +222,13 @@ def cli(
     ctx.color = real_color
     verbosity = verbose - quiet
     setup_logging(verbosity=verbosity, color=real_color)
+    if hide_token is None:
+        hide_token_bool = True
+    else:
+        if not trace:
+            option = "--hide-token" if hide_token else "--no-hide-token"
+            raise click.UsageError(f"{option} requires --trace")
+        hide_token_bool = hide_token
     root = Root(
         verbosity=verbosity,
         color=real_color,
@@ -220,6 +238,7 @@ def cli(
         network_timeout=network_timeout,
         config_path=Path(neuromation_config),
         trace=trace,
+        trace_hide_token=hide_token_bool,
     )
     ctx.obj = root
     if not ctx.invoked_subcommand:
--- a/neuromation/cli/main.py
+++ b/neuromation/cli/main.py
@@ -175,7 +175,6 @@ def print_options(
 )
 @click.option(
     "--trace",
-    default=None,
     is_flag=True,
     help="Trace sent HTTP requests and received replies to stderr.",
 )
--- a/neuromation/cli/root.py
+++ b/neuromation/cli/root.py
@@ -22,7 +22,7 @@ TEXT_TYPE = ("application/json", "text")
 
 HEADER_TOKEN_PATTERNS = [
     re.compile(rf"({auth_scheme})\s+([^ ]+\.[^ ]+\.[^ ]+)")
-    for auth_scheme in ("Bearer", "Basic")
+    for auth_scheme in ("Bearer", "Basic", "Digest", "Mutual")
 ]
 
 
--- a/neuromation/cli/root.py
+++ b/neuromation/cli/root.py
@@ -203,7 +203,9 @@ class Root:
         return text
 
     def _sanitize_token(self, token: str, tail_len: int = 5) -> str:
-        if not (0 < tail_len < len(token) // 2):
+        assert tail_len > 0, "invalid tail length"
+        # at least a third part of the token should be hidden
+        if tail_len >= len(token) // 3:
             return f"<hidden {len(token)} chars>"
 
         hidden = f"<hidden {len(token) - tail_len * 2} chars>"
--- a/neuromation/cli/root.py
+++ b/neuromation/cli/root.py
@@ -1,4 +1,5 @@
 import logging
+import re
 from dataclasses import dataclass
 from http.cookies import Morsel  # noqa
 from pathlib import Path
@@ -18,6 +19,11 @@ log = logging.getLogger(__name__)
 
 
 TEXT_TYPE = ("application/json", "text")
+
+HEADER_TOKEN_PATTERNS = [
+    re.compile(rf"({auth_scheme})\s+([^ ]+\.[^ ]+\.[^ ]+)")
+    for auth_scheme in ("Bearer", "Basic")
+]
 
 
 @dataclass
@@ -141,7 +147,7 @@ class Root:
             path += "?" + data.url.raw_query_string
         lines = [f"> {data.method} {path} HTTP/1.1"]
         for key, val in data.headers.items():
-            lines.append(f"> {key}: {val}")
+            lines.append(f"> {key}: {self._sanitize_header_value(val)}")
         lines.append("> ")
         self._print_debug(lines)
 
@@ -187,3 +193,8 @@ class Root:
             for line in self._process_chunk(chunk, context.response_printable)
         ]
         self._print_debug(lines)
+
+    def _sanitize_header_value(self, text: str) -> str:
+        for pattern in HEADER_TOKEN_PATTERNS:
+            text = pattern.sub(r"\1 <token>", text)
+        return text
--- a/neuromation/cli/root.py
+++ b/neuromation/cli/root.py
@@ -202,8 +202,8 @@ class Root:
             text = text.replace(token, token_safe)
         return text
 
-    def _sanitize_token(self, token: str, tail_len: int = 5) -> str:
-        assert tail_len > 0, "invalid tail length"
+    def _sanitize_token(self, token: str) -> str:
+        tail_len: int = 5
         # at least a third part of the token should be hidden
         if tail_len >= len(token) // 3:
             return f"<hidden {len(token)} chars>"
--- a/neuromation/cli/root.py
+++ b/neuromation/cli/root.py
@@ -203,9 +203,10 @@ class Root:
         return text
 
     def _sanitize_token(self, token: str, tail_len: int = 5) -> str:
-        assert 0 < tail_len, "tail too short"
-        assert tail_len < len(token) // 2, "tail too long"
-        hidden = f"<hidden {len(token) - tail_len*2} chars>"
+        if not (0 < tail_len < len(token) // 2):
+            return f"<hidden {len(token)} chars>"
+
+        hidden = f"<hidden {len(token) - tail_len * 2} chars>"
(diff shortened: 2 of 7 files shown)


ayushkovskiy:
I want this PR to be merged before we make a new release of `neuro` with support of `--trace`.

neuromation/cli/root.py
```diff
@@ -19,6 +20,11 @@
 
 TEXT_TYPE = ("application/json", "text")
 
+HEADER_TOKEN_PATTERNS = [
+    re.compile(rf"({auth_scheme})\s+([^ ]+\.[^ ]+\.[^ ]+)")
+    for auth_scheme in ("Bearer", "Basic", "Digest", "Mutual")
```
ayushkovskiy: I took the most popular auth types. Should we include others ([mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#Authentication_schemes) recommends also `HOBA`, `AWS4-HMAC-SHA256`.). What comes to other auth schemes: 
- `Bearer`: we use it
- `Basic`, `Digest` -- we might use it one day and thus current problem will be solved for these cases.  However, for sake of performance, we can remove these so-far-unsopported cases for now.
- `Mutual` -- as far as I know, we *will* use mutual authentication between client and server in future.

neuromation/cli/root.py
```diff
@@ -19,6 +20,11 @@
 
 TEXT_TYPE = ("application/json", "text")
 
+HEADER_TOKEN_PATTERNS = [
+    re.compile(rf"({auth_scheme})\s+([^ ]+\.[^ ]+\.[^ ]+)")
```
ayushkovskiy: We also can simplify the patterns not to know about JWT token format: `re.compile(rf"({auth_scheme})\s+([^ ]+")`

ayushkovskiy:
Still, my concern is user's security. So far, being in early stage of development, we haven't been carrying much on this aspect, however, IMO we should, especially with *the only authentication data the user has*. Token, being printed to the console, is registered to its history. Also, I consider the option `--trace` being unsafe to use as a debugging information we ask the users to provide in bug trackers as it will leak their secret information. I think, if the user wants their token to be printed, they can use the command `neuro config show-token`, otherwise the token should not be printed.

Anyways, I agree with a decision that is supported by common practices. Otherwise, both arguments are not sound.

anayden:
I don't think having full user token in output is safe. User may (and most likely eventually will) put their `--trace` output in public github issue or on pastebin.

Also I don't see much value for us to know the token.
In my experience from previous projects, passwords/tokens are always masked in the output, at least partially, unless explicitly config variable is set to not mask them.

asvetlov:
Well, partial masking sounds much better.
Also, if we do masking I'd like to have an option to disable the mask.
For example, I want to dig into potential problem by reproducing the failed operation with `curl`. All sent data is needed in this case.

ayushkovskiy:
> Well, partial masking sounds much better.

@asvetlov do you mean: `Authorization: Bearer eyJhbGciOiJI<hidden>MeJf36POk6yJV_adQssw5c`?

> if we do masking I'd like to have an option to disable the mask.

Agree, this option is necessary. How about: `neuro run --trace --no-hide-token ...` (or `--show-token`)?

asvetlov:
Please use `"--show-token/--hide-token", default=False` click spec.
Raise `click.BadOptionUsage` if `show_token and not trace`.

Regarding the output pattern: something like `Authorization: Bearer eyJ<hidden 42 chars>yfQ` would be good. First and last `N` chars or the token can help to make sure that the sent token looks correct. 
The exact `N` value is up to you, the whole line should fit in, e.g. 50-70 characters to never wrap it.

ayushkovskiy:
> Please use "--show-token/--hide-token", default=False click spec.

For me personally, `default=False` seems to be ambiguous: `False` to which of `show` or `hide`? How about `--hide-token/--no-hide-token, default=True`?

asvetlov:
`--no-hide-token/--no-no-hide-token`? I prefer shorter names.
For options, one or two words usually is enough for naming.
The default is never shown to a user in this case but specifies what flag is used by default.
Even `None` as *not set* is appropriate; completely internal rules for working with click API.

neuromation/cli/root.py
```diff
@@ -187,3 +195,19 @@ def _process_chunk(self, chunk: bytes, printable: bool) -> List[str]:
...
         self._print_debug(lines)
+
+    def _sanitize_header_value(self, text: str) -> str:
+        for token in self._find_all_tokens(text):
+            token_safe = self._sanitize_token(token)
+            text = text.replace(token, token_safe)
+        return text
+
+    def _sanitize_token(self, token: str, tail_len: int = 5) -> str:
+        assert 0 < tail_len, "tail too short"
```
asvetlov: The assertion is not appropriate here: I don't want to crash a program if the token is short.

tests/e2e/conftest.py
```diff
@@ -383,6 +383,7 @@ def run_cli(
         arguments: List[str],
         *,
         verbosity: int = 0,
+        raise_for_returncode: bool = True,
```
asvetlov: Please don't add the argument.
Catching `subprocess.CalledProcessError` can be used instead.

tests/e2e/test_e2e.py
```diff
@@ -32,3 +32,41 @@ def test_print_config_token(helper: Helper) -> None:
     captured = helper.run_cli(["config", "show-token"])
     assert not captured.err
     assert captured.out  # some secure information was printed
+
+
+@pytest.mark.e2e
+def test_root_trace_hide_token_default_true(helper: Helper) -> None:
```
asvetlov: Please drop e2e tests.
CLI tests are enough, the time cost of e2e is too high; let's no pessimize the test execution time if not necessary.

tests/e2e/conftest.py
ayushkovskiy: my motivation to change the `helper` was that when one catches `subprocess.CalledProcessError` via `pytest.raises`, there's no way that I know to assert stderr (`pytest.raises(..., match="text")` will match `text` only to stdout, however the HTTP trace is printed to stderr).

ayushkovskiy: however, if we drop e2e tests as it's proposed in another comment, we don't need this functionality at all.

ayushkovskiy: though I'm still curious how to assert on stderr with pytest if an exception was thrown.

tests/e2e/test_e2e.py
ayushkovskiy: Should we replace these e2e with a unit test on `def cli()` or just believe it works as intended?

neuromation/cli/root.py
ayushkovskiy: ok I agree these assertions shouldn't abort the execution, thank you.

tests/e2e/conftest.py
asvetlov: RTFD, dude: https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError

tests/e2e/test_e2e.py
asvetlov: I don't care

neuromation/cli/main.py
```diff
@@ -178,6 +178,16 @@ def print_options(
     is_flag=True,
     help="Trace sent HTTP requests and received replies to stderr.",
 )
+@click.option(
+    "--hide-token/--no-hide-token",
```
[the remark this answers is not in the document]
ayushkovskiy: No, it can be `reveal-token`, `demonstrate-token`,  `unfold-token`, `expose-token`,  `present-token`, `unhide-token` at the end. In boolean logic of `click`, the opposite to `--hide-token` is `--no-hide-token`, thus the user can easily understand what is the default option `True`.

Note also that `--no-no-hide-token` is an invalid option:
```
$ n --no-no-hide-token --trace ls
Usage: neuro [OPTIONS] COMMAND [ARGS]...
Try "neuro --help" for help.

Error: no such option: --no-no-hide-token
```

> The default is never shown. <...> Even None as not set is appropriate.

Thank you for this comment, I added the option `show_default=True` and changed the default `default=True` without changing the logic.

tests/e2e/test_e2e.py
ayushkovskiy: I have dropped two e2e tests that check that options `--hide-token` and `--no-hide-token` require the option `--trace` as this case is minor in terms of importance, and two other e2e tests that check this functionality if the option  `--hide-token` or `--no-hide-token` was specified explicitly. 

The only test I want to keep is `test_root_trace_hide_token_default_true` because it checks the use-case I worry about: when the user runs `neuro --trace <command>` and their token is not revealed.

neuromation/cli/main.py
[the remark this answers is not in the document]
ayushkovskiy: > I added the option show_default=True and changed the default default=True without changing the logic

This approach was erroneous. I solved this problem in the same way as it's solved for `"--http-auth/--no-http-auth"`: `is_flag=True`, `default=None` and the default value is manually written to the help message.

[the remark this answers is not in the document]
asvetlov: ok

neuromation/cli/root.py
ayushkovskiy: not the case anymore.

neuromation/cli/root.py
```diff
@@ -187,3 +195,20 @@ def _process_chunk(self, chunk: bytes, printable: bool) -> List[str]:
...
         self._print_debug(lines)
+
+    def _sanitize_header_value(self, text: str) -> str:
+        for token in self._find_all_tokens(text):
+            token_safe = self._sanitize_token(token)
+            text = text.replace(token, token_safe)
+        return text
+
+    def _sanitize_token(self, token: str, tail_len: int = 5) -> str:
+        if not (0 < tail_len < len(token) // 2):
```
asvetlov: Why is `0` here? Do you suppose to support negative lengths?
I think the condition can be significantly simplified.

neuromation/cli/main.py
```diff
@@ -178,6 +178,16 @@ def print_options(
...
+    is_flag=True,
+    default=None,
+    help=(
+        "Prevent user's token sent in HTTP headers from being "
+        "printed out to stderr during HTTP tracing. Can be used only "
+        "together with option '--trace'  [default: True]"
```
asvetlov: `default: True` is meaningless for flags; I have no idea what option matches to `True` or `False` on the reading.

tests/cli/test_root.py
```diff
@@ -50,3 +50,52 @@ def test_resource_presets_uninitialized(root_uninitialized: Root) -> None:
...
+
+class TestTokenSanitization:
+    @pytest.mark.parametrize(
+        "auth", ["Bearer", "Basic", "Digest", "Mutual"],
+    )
+    def test_sanitize_header_value_single_token(
+        self, root_uninitialized: Root, auth: str
+    ) -> None:
+        line = f"foo Authentication: {auth} eyJhbGciOiJI.eyJzdW0NTY3.SfKxwRJ_SsM bar"
+        expected = f"foo Authentication: {auth} eyJhb<hidden 26 chars>J_SsM bar"
```
asvetlov: `_sanitize_header_value` is for the value by definition, a test for `key: value` string is misleading.

The same below.

tests/e2e/conftest.py
asvetlov: So, please revert the file entirely.

ayushkovskiy: ok

ayushkovskiy: sorry forgot to revert the helpers as well.

neuromation/cli/root.py
ayushkovskiy: `tail_len` is the parameter set up manually (currently, only in signature of `_sanitize_token`, in future it's possible to expose it further). Thus, we need to assert that it's a valid length. I agree it'll be more clear and idiomatic if the funciton fails on non-positive values of `tail_len`.

neuromation/cli/main.py
ayushkovskiy: 1. it is much more clear than `--hide-token` (as True) and `--show-token` (as False).
2. if still it is meaningless, let's then fix all other similar places:

https://github.com/neuromation/platform-client-python/blob/72416311c23a0eb6c0e99053c53268d5e1efc977/neuromation/cli/job.py#L151-L158

https://github.com/neuromation/platform-client-python/blob/72416311c23a0eb6c0e99053c53268d5e1efc977/neuromation/cli/job.py#L162-L167

https://github.com/neuromation/platform-client-python/blob/72416311c23a0eb6c0e99053c53268d5e1efc977/neuromation/cli/job.py#L168-L174

https://github.com/neuromation/platform-client-python/blob/72416311c23a0eb6c0e99053c53268d5e1efc977/neuromation/cli/job.py#L222-L227

https://github.com/neuromation/platform-client-python/blob/72416311c23a0eb6c0e99053c53268d5e1efc977/neuromation/cli/job.py#L228-L233

https://github.com/neuromation/platform-client-python/blob/72416311c23a0eb6c0e99053c53268d5e1efc977/neuromation/cli/job.py#L315-L321

https://github.com/neuromation/platform-client-python/blob/72416311c23a0eb6c0e99053c53268d5e1efc977/neuromation/cli/storage.py#L65-L71

https://github.com/neuromation/platform-client-python/blob/72416311c23a0eb6c0e99053c53268d5e1efc977/neuromation/cli/storage.py#L222-L228

If not, let's move on to do more important things.

tests/cli/test_root.py
ayushkovskiy: my bad, fixing.

neuromation/cli/root.py
asvetlov: Thanks for the reminder.
The idiomatic approach is don't add the unused functionality.
Drop `tail_len` from the method parameter, convert it into `tail_len = 5` local variable.

ayushkovskiy:
Added a changelog

neuromation/cli/main.py
asvetlov: Honestly, I wanted to fix the places mentioned by you eventually.
"Show progress, on by default" from the last snippet is a good text IMHO.

The fix of existing options worth another PR, adding a new option with text that should be fixed from the very beginning is suboptimal.

ayushkovskiy: fixed message to `... On by default.`

ayushkovskiy:
e2e jobs fail because of the problems on dev cluster, let's wait

PR #1164, opened 2019-11-13
Make copying progress less blinking.

### LINKED ISSUE #1163: Files jump while `neuro cp -r`
See https://asciinema.org/a/xpl1M3ObqLqOJzsBBhOg2Dex8

I have received multiple complains of the users who try `neuro cp -r` for the first time: for them, file names blinking on the screen looks like a bug. Perhaps, it's better not to re-write lines in terminal, but just print them on new line (as, for exapmle, zip does: https://asciinema.org/a/ypszHC9MvRXYaXCqO7I0l02vc)

serhiy-storchaka:
* Decrease the number of simultaneous copied files to 20.
* Increase the height of the progress screen to 25.

Fixes #1163.


serhiy-storchaka:
I tested earlier -- decrease the number of simultaneous copied files to 10 significantly affects performance. With 20 the negative effect is much smaller.

serhiy-storchaka:
We need also to limit the frequency of updating the screen.

ayushkovskiy:
From my point of view, this PR does not fix the original problem of issue https://github.com/neuromation/platform-client-python/issues/1163 (blinking screen), see https://github.com/neuromation/platform-client-python/issues/1163#issuecomment-554081773

Issue #897, opened 2019-07-10
Check whether a docker engine is available on neuro image push


dalazx:
![Screenshot](https://user-images.githubusercontent.com/1091425/60980431-232d1380-a33d-11e9-87c2-4909e9cba82b.png)

cc @mariyadavydova

asvetlov:
done

PR #1169, opened 2019-11-15
Better Docker api connection error handling

shagren:
Alternative solution for  #897
Just show friendly message instead ping docker engine

On my system when docker service stopped unix socket still exists. And with this PR i see now:
> ERROR: Docker API error: Cannot connect to Docker engine via /run/docker.sock, Connection refused

it was before: 
>ERROR: Connection error (Cannot connect to host localhost:None ssl:default [Connection refused])

For remote host it will be:
>ERROR: Docker API error: Cannot connect to Docker engine via tcp://1.2.3.4:123, Connect call failed ('1.2.3.4', 123)

instead
>ERROR: Connection error (Cannot connect to host 1.2.3.4:123 ssl:default [Connect call failed ('1.2.3.4', 123)])

### DIFF
--- a/neuromation/api/images.py
+++ b/neuromation/api/images.py
@@ -6,6 +6,7 @@ from typing import Any, Dict, List, Optional, Set
 import aiodocker
 import aiohttp
 from aiodocker.exceptions import DockerError
+from aiohttp.client_exceptions import ClientConnectorError
 from yarl import URL
 
 from .abc import (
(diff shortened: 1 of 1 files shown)


asvetlov:
The change makes sense.
Would you consider patching the upstream instead?

asvetlov:
Fixed by aiodocker's master

PR #1184, opened 2019-11-21
Drop annoying git hooks

asvetlov:


Makefile
```diff
@@ -1,20 +1,12 @@
-DEVPI_URL ?= "https://$(DEVPI_USER):$(DEVPI_PASS)@$(DEVPI_HOST)/$(DEVPI_USER)"
-
 SHELL := /bin/bash
 BRANCH := $(shell git branch | grep \* | cut -d ' ' -f2)
 
 MYPY_CACHE_DIR=.mypy_cache/$(shell md5sum setup.py | awk '{print $$1}')-$(shell find requirements -type f -exec md5sum {} \; | sort -k 2 | md5sum | awk '{print $$1}')
```
bmkz: Let's try to remove this hack too?

asvetlov: I thought of it. 
Well, why not. 
Mypy had problems with the cache invalidation but now everything is fixed AFAIK.
At least we use mypy without this setting on other projects successfully.

PR #1195, opened 2019-11-25
Implement `config get-clusters` command

### LINKED ISSUE #1177: Add `neuro config get-clusters`
from https://docs.google.com/document/d/1A9STpF-9vlaVAaCisCqo-o7NU-Kikc9aFsEHOwImuKY/edit#

**neuro config get-clusters**
Pulls the ready-to-use clusters available to the user;
Outputs the list of clusters;
Nothing is saved locally.

Once https://github.com/neuromation/platform-api/issues/966 is ready, we could reuse some existing code to request `GET /api/v1/config`, parse the `clusters` key, and output the rendered list of ready-to-use clusters.

asvetlov:
Fixes #1177

### DIFF
--- a/neuromation/cli/config.py
+++ b/neuromation/cli/config.py
@@ -17,11 +17,12 @@ from neuromation.api import (
     login_with_token as api_login_with_token,
     logout as api_logout,
 )
-from neuromation.cli.formatters.config import QuotaInfoFormatter
+from neuromation.api.server_cfg import get_server_config
+from neuromation.cli.formatters.config import ClustersFormatter, QuotaInfoFormatter
 
 from .formatters import ConfigFormatter
 from .root import Root
-from .utils import async_cmd, command, group
+from .utils import async_cmd, command, group, pager_maybe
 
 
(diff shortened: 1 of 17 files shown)


neuromation/cli/config.py
```diff
@@ -198,12 +199,28 @@ def config() -> None:
...
+async def get_clusters(root: Root) -> None:
+    """
+    Fetch and display the list of available clusters from the Neuro Platform.
+
+    """
+
+    config = await get_server_config(root.client._session, root.url, root.auth)
+    fmt = ClustersFormatter()
+    pager_maybe(
+        fmt(config.clusters, config.cluster_config.name), root.tty, root.terminal_size
```
dalazx: I think here instead of `config.cluster_config.name` we should use the name of the cluster stored in the user's local config.

asvetlov: Makes sense.
The value doesn't exist at the moment though, I'll add it on implementing `select-cluster` command.

PR #1217, opened 2019-12-03
Select cluster

asvetlov:


neuromation/api/jobs.py
```diff
@@ -160,44 +163,51 @@ def __init__(self, core: _Core, config: _Config, parse: Parser) -> None:
...
+        url = self._config.api_url / "jobs"
         params: MultiDict[str] = MultiDict()
         for status in statuses:
             params.add("status", status.value)
         if name:
             params.add("name", name)
         for owner in owners:
             params.add("owner", owner)
-        async with self._core.request("GET", url, params=params) as resp:
+        params["cluster_name"] = self._config.cluster_name
```
dalazx: should we consider making this optional? `neuro ps` renders jobs' cluster names already if I'm not mistaken.

asvetlov:
The code is ready for review; I want to add tests and update API documentation before the merge.

neuromation/api/jobs.py
asvetlov: Thanks for the excellent question!
The PR selects `cluster['name']` if the name is specified, otherwise `'default'` is used for old deployments.
Is `'default'` good enough or it will be rejected by trafarets on servers?

dalazx: `'default'` is a good transitory option.
on `POST /api/v1/jobs` we validate the passed cluster name (if any, otherwise we fall back to `default`) against the clusters available to the user.
on `GET /api/v1/jobs` we just make sure that `cluster_name` is a valid name.

there's a UX part of the question: a user performs the following operations:
```
neuro run ...
neuro config switch-cluster ...
neuro run ...
```
should we output both jobs on
```
neuro ps
```
or only the ones running on the active cluster?

neuromation/api/server_cfg.py
```diff
@@ -88,17 +65,37 @@ def _parse_cluster_config(payload: Dict[str, Any]) -> _ClusterConfig:
...
+    else:
+        # old server without multiple clusters support
+        cluster = _parse_cluster_config(payload)
+        if (
+            cluster.registry_url
+            and cluster.storage_url
+            and cluster.users_url
+            and cluster.monitoring_url
+            and cluster.resource_presets
+        ):
```
dalazx:
```python
if _is_cluster_config_initialized(cluster):
```

neuromation/api/config.py
```diff
@@ -102,15 +113,183 @@ def to_config(self) -> Dict[str, Any]:
...
+        return name
+
+    async def fetch(self) -> None:
+        server_config = await get_server_config(
+            self._core._session,
+            self._config_data.url,
+            self._config_data.auth_token.token,
+        )
+        if self.cluster_name not in server_config.clusters:
+            raise RuntimeError(
```
dalazx: if a user was removed from a cluster, how would they recover from such a situation?

neuromation/cli/config.py
```diff
@@ -203,24 +203,75 @@ def config() -> None:
...
+        "The current cluster is " + click.style(real_cluster_name, underline=True)
+    )
+
+
+async def prompt_cluster(client: Client) -> str:
+    clusters = client.config.clusters
+    while True:
+        fmt = ClustersFormatter()
+        click.echo("\n".join(fmt(clusters.values(), client.config.cluster_name)))
+        answer = input(f"Select cluster to switch [{client.config.cluster_name}]: ")
```
dalazx: if a user hits return right away, should we fall back to `[{client.config.cluster_name}]`?

tests/api/test_config.py
```diff
@@ -1,128 +1,161 @@
...
+    JSON = {
+        "auth_url": auth_url,
+        "token_url": token_url,
+        "client_id": client_id,
+        "audience": audience,
+        "headless_callback_url": headless_callback_url,
+        "success_redirect_url": success_redirect_url,
+        "clusters": [
+            {
+                "registry_url": registry_url,
```
dalazx: minor. `"name"` is missing - it's always present in the response.

neuromation/api/users.py
```diff
@@ -48,9 +52,11 @@ def __init__(self, core: _Core) -> None:
     async def get_shares(
         self, user: str, scheme: Optional[str] = None
     ) -> Sequence[Share]:
-        url = URL(f"users/{user}/permissions/shared")
+        url = self._config.api_url / "users" / user / "permissions" / "shared"
```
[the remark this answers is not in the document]
shagren: I mean why not `self._config.api_url / "users" / user / "permissions/shared"`

neuromation/api/jobs.py
asvetlov: I have no idea. Now, only the current cluster's jobs are shown.  
Like in Kubernetes, I see resources from the selected context only.
I suggest keep it until people complain.

neuromation/api/config.py
asvetlov: Perhaps user's config is not relevant in this case, the exception text suggests re-login as the last resort.

neuromation/api/users.py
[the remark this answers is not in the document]
asvetlov: Result is the same.

neuromation/api/server_cfg.py
```diff
@@ -88,17 +65,37 @@ def _parse_cluster_config(payload: Dict[str, Any]) -> _ClusterConfig:
...
 
 
+def _parse_clusters(payload: Dict[str, Any]) -> Dict[str, ClusterConfig]:
+    ret: Dict[str, ClusterConfig] = {}
+    if "clusters" in payload:
+        for item in payload["clusters"]:
+            cluster = _parse_cluster_config(item)
+            ret[cluster.name] = cluster
+    else:
+        # old server without multiple clusters support
```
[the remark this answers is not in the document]
asvetlov: #1218

asvetlov:
Multi-file configuration makes sense.
I prefer to address it in another PR, this one is too big already.

neuromation/api/config.py
```diff
@@ -102,15 +113,183 @@ def to_config(self) -> Dict[str, Any]:
...
+        self._config_data = replace(self._config_data, cluster_name=name)
+        self._save(self._config_data, self._path)
+
+    @property
+    def api_url(self) -> URL:
+        return self._config_data.url
+
+    @property
+    def monitoring_url(self) -> URL:
+        cluster = self._config_data.clusters[self._config_data.cluster_name]
```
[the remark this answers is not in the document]
asvetlov: It is not strictly necessary.
I was striving to keep the public API as minimal as possible, the proposed property can be used in 4 short lines of the source code.
I'm not against the idea, maybe later the property will make more sense. Let's see how things will go

neuromation/cli/config.py
```diff
@@ -203,24 +203,77 @@ def config() -> None:
...
+        answer = answer.strip()
+        if answer is None:
```
dalazx: seems like this will never be the case since `type(answer) is str`.

dalazx:
if we always do `fetch` to retrieve the latest available clusters on `get-clusters` and `switch-cluster`, why do we need to store all of the available clusters in the config files, not only the active one? is there a plan to avoid `fetch`ing and rely on the locally-stored data at some point ?

neuromation/cli/config.py
asvetlov: Good catch!
I will fix it soon.

asvetlov:
Nice question!
Saving all cluster configurations cost nothing, so my initial thought was: "why not"?
Having clusters in saved config makes API a little cleaner: `client.config.clusters` is available without explicit `await client.config.fetch()` etc.
Storing the current cluster data is possible, but I don't see a reason for this restriction.
I wanted to save the current only when edited a proposal document; on implementing the feature, I see we can keep everything.

Regarding CLI: I think the fetching of available clusters before the switch makes the system a little safer; by this, we avoid the situation when a user switches to a cluster that doesn't exist anymore but was present at the time of the last `neuro config get-clusters` call.

neuromation/cli/config.py
asvetlov: Done. Writing a test for the helper function.

asvetlov:
Thanks, @dalazx 
Sure, the system will need a love for tuning.
What I really like now is introducing the public config class, it is a big step forward.

PR #1226, opened 2019-12-05
Rename .nmrc to .neuro

asvetlov:


asvetlov:
I've updated `CHANGELOG.D/1183.feature` to reflect the new naming.

PR #1233, opened 2019-12-09
Api changes: quota per cluster.

### LINKED ISSUE #1179: Adjust `neuro config show-quota` to show the quota for the active cluster
Once https://github.com/neuromation/platform-client-python/issues/1175 and https://github.com/neuromation/platform-api/issues/992 are ready, we will need to look up the cluster stats by `cluster_name` currently stored in `.nmrc`.

shagren:
closes #1179

### DIFF
--- a/neuromation/api/quota.py
+++ b/neuromation/api/quota.py
@@ -9,23 +9,26 @@ from neuromation.api.utils import NoPublicConstructor
 @dataclass(frozen=True)
 class _QuotaInfo:
     cluster_name: str
-    cpu_time_spent: float
-    cpu_time_limit: float
-    gpu_time_spent: float
-    gpu_time_limit: float
+    # All time in seconds
+    cpu_time_spent: int
+    cpu_time_limit: Optional[int]
+    gpu_time_spent: int
+    gpu_time_limit: Optional[int]
 
     @property
-    def cpu_time_left(self) -> float:
+    def cpu_time_left(self) -> Optional[int]:
         return self._get_remaining_time(self.cpu_time_spent, self.cpu_time_limit)
 
     @property
-    def gpu_time_left(self) -> float:
+    def gpu_time_left(self) -> Optional[int]:
         return self._get_remaining_time(self.gpu_time_spent, self.gpu_time_limit)
 
-    def _get_remaining_time(self, time_spent: float, time_limit: float) -> float:
-        if time_limit > time_spent:
-            return time_limit - time_spent
-        return 0.0
+    def _get_remaining_time(
+        self, time_spent: int, time_limit: Optional[int]
+    ) -> Optional[int]:
+        if time_limit is None:
+            return None
+        return max(time_limit - time_spent, 0)
 
 
 class _Quota(metaclass=NoPublicConstructor):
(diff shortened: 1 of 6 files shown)


neuromation/api/quota.py
```diff
@@ -8,7 +8,6 @@
 
 @dataclass(frozen=True)
 class _QuotaInfo:
-    name: str
```
dalazx: we could still store a cluster name here.

neuromation/api/quota.py
```diff
@@ -42,17 +41,23 @@ def __init__(self, core: _Core, config: Config) -> None:
...
-        cpu_time_spent=spent_cpu,
-        cpu_time_limit=limit_cpu,
-    )
+def _quota_info_from_api(payload: Dict[str, Any]) -> Dict[str, _QuotaInfo]:
+    clusters = payload["clusters"]
+    ret: Dict[str, _QuotaInfo] = {}
+    for cluster_payload in clusters:
+        cluster_name = cluster_payload["name"]
+        jobs_payload = cluster_payload["jobs"]
+        quota_payload = cluster_payload.get("quota", {})
```
dalazx: looks like `quota` is always present in the response: https://github.com/neuromation/platform-api/blob/115c3f07e02356aec14fb473f71b799095f934d7/platform_api/handlers/stats_handler.py#L96

dalazx: but might be `{}` an empty dict, yep.

neuromation/api/quota.py
shagren: Why we need name here?
Dict key will be more useful for end-user

dalazx: `QuotaInfoFormatter` could unambiguously output the cluster name for example.

shagren: added

neuromation/api/quota.py
shagren: There was test for missed quota before. Then I add this case. Can I safetly remove or better to leave for future?

dalazx: we can leave as is. up to you.

neuromation/cli/config.py
```diff
@@ -54,11 +54,18 @@ def config() -> None:
...
     """
-    Print quota and remaining computation time.
+    Print quota and remaining computation time for active cluster.
     """
-    quota = await root.client._quota.get(user)
+    quotas = await root.client._quota.get(user)
+    cluster_name = root.client.config.cluster_name
+    if cluster_name not in quotas:
+        raise ValueError(
+            f"No quota information available for cluster {cluster_name}\n"
```
serhiy-storchaka:
```suggestion
            f"No quota information available for cluster {cluster_name}.\n"
```

neuromation/api/quota.py
```diff
@@ -42,17 +42,24 @@ def __init__(self, core: _Core, config: Config) -> None:
...
+        ret[cluster_name] = _QuotaInfo(
+            cluster_name=cluster_name,
+            cpu_time_spent=float(
+                int(jobs_payload["total_non_gpu_run_time_minutes"]) * 60
```
serhiy-storchaka: This was in the current code, but why do we need to call `int()` and `float()`? Are not all these values already integers?

shagren: Good catch! Will change soon

shagren: @serhiy-storchaka , can you check latest implementation?

neuromation/api/quota.py
```diff
@@ -42,17 +45,23 @@ def __init__(self, core: _Core, config: Config) -> None:
...
+        quota_payload = cluster_payload.get("quota", {})
+        cpu_time_spent = jobs_payload["total_non_gpu_run_time_minutes"]
+        gpu_time_spent = jobs_payload["total_gpu_run_time_minutes"]
+        cpu_time_limit = quota_payload.get("total_non_gpu_run_time_minutes", None)
+        gpu_time_limit = quota_payload.get("total_gpu_run_time_minutes", None)
+
+        ret[cluster_name] = _QuotaInfo(
+            cluster_name=cluster_name,
+            cpu_time_spent=cpu_time_spent * 60,
+            cpu_time_limit=cpu_time_limit * 60 if cpu_time_limit else None,
```
serhiy-storchaka: What if `cpu_time_limit` is 0?

neuromation/cli/formatters/config.py
```diff
@@ -47,22 +47,22 @@ def __call__(self, quota: _QuotaInfo) -> str:
...
         else:
             quota_str += self.QUOTA_NOT_SET
         return f"{spent_str} ({quota_str})"
 
-    def _format_time(self, total_seconds: float) -> str:
+    def _format_time(self, total_seconds: int) -> str:
         # Since API for `GET /stats/users/{name}` returns time in minutes,
         #  we need to display it in minutes as well.
-        total_minutes = int(total_seconds // 60)
+        total_minutes = total_seconds // 60
```
serhiy-storchaka: It is not related to this PR, but I think the code below can be simplified.
```python
hours, minutes = divmod(total_minutes, 60)
return f"{hours:02d}h {minutes:02d}m"
```

asvetlov:
What is the reason for float->int replacement?

serhiy-storchaka:
The reason is that all values are integers and calling `float()` to convert from integers from the Web API is boring. But AFAIK `int` is considered a subtype of `float` in type annotations, so we can keep `float` in annotations and set fields to `int` value (or `float('inf'`)).

I have no preferences, both approaches (`float` and `Optional[int]`) work to me.

asvetlov:
Python time has 2 representations usually: it is either one of `datetime` classes or `float` as returned by `time.time()` and others. `time.struct_time` has too cumbersome API, it's usage is limited from my experience.

That's why the previous version used `float`. I guess people are very familiar with datetime-as-float types.
Quota API is private now but I expect it will become public eventually after a stabilization period. This is the reason why I do care about used types.

neuromation/cli/formatters/config.py
shagren: Sure, changed

shagren:
@asvetlov, thnx for explanation. Reverted to float back.

shagren:
@asvetlov can we merge red? Like as errors not related to this PR

PR #1234, opened 2019-12-09
Added get-cluster-users command

tvoinarovskyi:


asvetlov:
There is a possibility to write e2e for get-* commands, the test can run the corresponding command without actual checking for the result.
Tests for changing the state is more challenging, I doubt if we can test add-cluster for example.
Maybe fake server tests for API are enough?

PR #1236, opened 2019-12-10
Add remaining cluster-user commands

### LINKED ISSUE #1221: Add `neuro admin add-cluster-user`
```
neuro admin add-cluster-user <cluster-name> <user-name> [role: admin|manager|user]
```

`user` is the default role - could be omitted.

```
curl -vv -H "Authorization: Bearer $TOKEN" -X POST -d '{"user_name": "user-name", "role": "user"}' https://dev.neu.ro/apis/admin/v1/clusters/$CNAME/users
```

https://docs.google.com/document/d/1A9STpF-9vlaVAaCisCqo-o7NU-Kikc9aFsEHOwImuKY/edit#

### LINKED ISSUE #1222: Add `neuro admin remove-cluster-user`
```
neuro admin remove-cluster-user <cluster-name> <user-name>
```

```
curl -vv -H "Authorization: Bearer $TOKEN" -X DELETE https://dev.neu.ro/apis/admin/v1/clusters/$CNAME/users/$UNAME

```

tvoinarovskyi:
closes #1222 
closes #1221

### DIFF
--- a/neuromation/api/admin.py
+++ b/neuromation/api/admin.py
@@ -56,7 +56,8 @@ class _Admin(metaclass=NoPublicConstructor):
         auth = await self._config._api_auth()
 
         async with self._core.request("DELETE", url, auth=auth) as resp:
-            await resp.text()
+            # No content response
+            await resp.release()
 
 
 def _cluster_user_from_api(payload: Dict[str, Any]) -> _ClusterUser:
--- a/neuromation/api/admin.py
+++ b/neuromation/api/admin.py
@@ -41,15 +41,13 @@ class _Admin(metaclass=NoPublicConstructor):
     async def add_cluster_user(
         self, cluster_name: str, user_name: str, role: str
     ) -> _ClusterUser:
-        role_type = _ClusterUserRoleType(role)
-        user = _ClusterUser(user_name, role_type)
         url = self._config.admin_url / "clusters" / cluster_name / "users"
         payload = {"user_name": user_name, "role": role}
         auth = await self._config._api_auth()
 
         async with self._core.request("POST", url, json=payload, auth=auth) as resp:
-            await resp.json()
-        return user
+            payload = await resp.json()
+            return _cluster_user_from_api(payload)
 
     async def remove_cluster_user(self, cluster_name: str, user_name: str) -> None:
         url = self._config.admin_url / "clusters" / cluster_name / "users" / user_name
--- a/neuromation/api/admin.py
+++ b/neuromation/api/admin.py
@@ -56,7 +56,7 @@ class _Admin(metaclass=NoPublicConstructor):
         auth = await self._config._api_auth()
 
         async with self._core.request("DELETE", url, auth=auth) as resp:
-            await resp.json()
+            await resp.text()
 
 
 def _cluster_user_from_api(payload: Dict[str, Any]) -> _ClusterUser:
--- a/neuromation/api/admin.py
+++ b/neuromation/api/admin.py
@@ -38,6 +38,19 @@ class _Admin(metaclass=NoPublicConstructor):
             res = await resp.json()
             return [_cluster_user_from_api(payload) for payload in res]
 
+    async def add_cluster_user(
+        self, cluster_name: str, user_name: str, role: str
+    ) -> _ClusterUser:
+        role_type = _ClusterUserRoleType(role)
+        user = _ClusterUser(user_name, role_type)
+        url = self._config.admin_url / "clusters" / cluster_name / "users"
+        payload = {"user_name": user_name, "role": role}
+        auth = await self._config._api_auth()
+
+        async with self._core.request("POST", url, json=payload, auth=auth) as resp:
+            await resp.json()
+        return user
+
 
 def _cluster_user_from_api(payload: Dict[str, Any]) -> _ClusterUser:
     return _ClusterUser(
--- a/neuromation/api/admin.py
+++ b/neuromation/api/admin.py
@@ -55,7 +55,7 @@ class _Admin(metaclass=NoPublicConstructor):
 
         async with self._core.request("DELETE", url, auth=auth) as resp:
             # No content response
-            await resp.release()
+            pass
 
 
 def _cluster_user_from_api(payload: Dict[str, Any]) -> _ClusterUser:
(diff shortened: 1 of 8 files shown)


neuromation/cli/admin.py
```diff
@@ -26,4 +26,31 @@ def admin() -> None:
...
+
+@command()
+@click.argument("cluster_name", required=True, type=str)
+@click.argument("user_name", required=True, type=str)
+@async_cmd()
+async def remove_cluster_user(root: Root, cluster_name: str, user_name: str) -> None:
+    """
+    Remove user access from the cluster
+    """
+    await root.client._admin.remove_cluster_user(cluster_name, user_name)
```
tvoinarovskyi: Should we print something to the user, like: 
Added user: Ivan[admin]
Removed user: Alex

asvetlov: I think yes.
Ideally, we should print texts as you had in examples in the normal mode and skip printing if `--quiet` is specified.
Adding ANSI escapes (bold, underline, etc.) can improve visual readability, see `neuro config show` and others for inspiration.

neuromation/api/admin.py
```diff
@@ -38,6 +38,26 @@ def __init__(self, core: _Core, config: Config) -> None:
...
+        self, cluster_name: str, user_name: str, role: str
+    ) -> _ClusterUser:
+        role_type = _ClusterUserRoleType(role)
+        user = _ClusterUser(user_name, role_type)
+        url = self._config.admin_url / "clusters" / cluster_name / "users"
+        payload = {"user_name": user_name, "role": role}
+        auth = await self._config._api_auth()
+
+        async with self._core.request("POST", url, json=payload, auth=auth) as resp:
+            await resp.json()
```
serhiy-storchaka: Loading and parsing a JSON is not needed if the value is ignored.

Or maybe use `_cluster_user_from_api()` for creating the returning value?

neuromation/cli/admin.py
```diff
@@ -26,4 +26,31 @@ def admin() -> None:
     )
 
 
+@command()
+@click.argument("cluster_name", required=True, type=str)
+@click.argument("user_name", required=True, type=str)
+@click.argument("role", required=False, default="user", type=str)
```
serhiy-storchaka: Maybe use `click.Choice`?

serhiy-storchaka:
Maybe add tests for errors? Removing non-existing user, adding already existing user?

neuromation/api/admin.py
tvoinarovskyi: Probably will do the 2nd one.

neuromation/cli/admin.py
```diff
@@ -38,6 +38,11 @@ def admin() -> None:
     Add user access to specified cluster with one of 3 roles: admin, manager or user
     """
     await root.client._admin.add_cluster_user(cluster_name, user_name, role)
+    if root.verbosity >= 0:
```
serhiy-storchaka: Or `not root.quiet`.

neuromation/cli/admin.py
tvoinarovskyi: Added echo calls with quiet support

neuromation/cli/admin.py
tvoinarovskyi: done

neuromation/cli/admin.py
tvoinarovskyi: done

neuromation/cli/admin.py
```diff
@@ -26,4 +28,46 @@ def admin() -> None:
     )
 
 
+@command()
+@click.argument("cluster_name", required=True, type=str)
+@click.argument("user_name", required=True, type=str)
+@click.argument(
+    "role",
+    required=False,
+    default="user",
```
dalazx:
```suggestion
    default= _ClusterUserRoleType.USER.value,
```
minor.

tvoinarovskyi:
> Maybe add tests for errors? Removing non-existing user, adding already existing user?

Added as e2e tests, as those don't alter the state of the cluster. Please give it another view.

neuromation/cli/admin.py
```diff
@@ -26,4 +28,47 @@ def admin() -> None:
...
+async def add_cluster_user(
+    root: Root, cluster_name: str, user_name: str, role: str
+) -> None:
+    """
+    Add user access to specified cluster with one of 3 roles: admin, manager or user
+    """
+    await root.client._admin.add_cluster_user(cluster_name, user_name, role)
+    if not root.quiet:
+        click.echo(
+            f"Added {click.style(user_name, bold=True)} to cluster "
```
serhiy-storchaka: Maybe use attributes of the result of `add_cluster_user()`: `user.user_name`, `user.role`? Just for the case if the server changes them.

neuromation/api/admin.py
```diff
@@ -38,6 +38,25 @@ def __init__(self, core: _Core, config: Config) -> None:
...
+            payload = await resp.json()
+            return _cluster_user_from_api(payload)
+
+    async def remove_cluster_user(self, cluster_name: str, user_name: str) -> None:
+        url = self._config.admin_url / "clusters" / cluster_name / "users" / user_name
+        auth = await self._config._api_auth()
+
+        async with self._core.request("DELETE", url, auth=auth) as resp:
+            # No content response
+            await resp.release()
```
asvetlov:
```suggestion
            pass
```
The response is released on exit from `async with`

neuromation/cli/admin.py
```diff
@@ -26,4 +28,47 @@ def admin() -> None:
     )
 
 
+@command()
+@click.argument("cluster_name", required=True, type=str)
+@click.argument("user_name", required=True, type=str)
+@click.argument(
```
asvetlov: That's interesting: `click` renders `ROLE` argument as mandatory according to README.

tvoinarovskyi: Yea and if I remove metavar it prints bad error messages:

```
Invalid value for "[[admin|manager|user]]": invalid choice: my_role. (choose from admin, manager, user)
```

Don't know which is better

serhiy-storchaka: What if set `metavar="[ROLE]"`?

tvoinarovskyi: Yea, at least that looks better

tvoinarovskyi:
@serhiy-storchaka Added test to check that we print values returned from the server, not from input params.

PR #1238, opened 2019-12-10
Fix tests run for mac OSX

tvoinarovskyi:
On OSX localhosts only corresponds to 127.0.0.1/32 by default

### DIFF
--- a/tests/api/test_config_factory.py
+++ b/tests/api/test_config_factory.py
@@ -1,7 +1,7 @@
 import shutil
 import sys
 from pathlib import Path
-from typing import Any, Dict
+from typing import Any, Callable, Dict
 from unittest import mock
 
 import aiohttp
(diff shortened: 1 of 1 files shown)


tests/api/test_config_factory.py
```diff
@@ -44,19 +44,26 @@ def config_dir(
...
 @pytest.fixture
-async def mock_for_login(aiohttp_server: _TestServerFactory, token: str) -> _TestServer:
+async def mock_for_login(
+    aiohttp_server: _TestServerFactory,
+    token: str,
+    aiohttp_unused_port: Callable[[], int],
+) -> _TestServer:
+
+    callback_urls = [
+        f"http://127.0.0.1:{aiohttp_unused_port()}",
```
serhiy-storchaka: Do repeated calls of `aiohttp_unused_port()` return different values every time?

asvetlov: It seems so

serhiy-storchaka: Seems this is not guaranteed. There is a 0.005% chance of getting non-unique values in three calls of `aiohttp_unused_port()`. Hope this will not break the tests.

serhiy-storchaka:
The default value for `_AuthConfig.callback_urls` is
```python
(
        URL("http://127.0.0.1:54540"),
        URL("http://127.0.0.1:54541"),
        URL("http://127.0.0.1:54542"),
    )
```
Should not it be changed or removed at all?

tvoinarovskyi:
Hmm, the failures do not seem related to thi PR scope.

asvetlov:
@serhiy-storchaka agree, defaults are always provided by the server.

asvetlov:
@tvoinarovskyi please try to re-run tests

tvoinarovskyi:
> @tvoinarovskyi please try to re-run tests

Does not help :(

asvetlov:
The error is: `neuromation.api.core.IllegalArgumentError: encoding with 'idna' codec failed (UnicodeError: label too long)`

It is not related to your work; I'm preparing a fix

PR #1241, opened 2019-12-11
Delay the output of the file copying progress.

serhiy-storchaka:


serhiy-storchaka:
It does not work yet.

PR #1242, opened 2019-12-11
Ignore invalid storage path when use --volume=ALL

asvetlov:


dalazx:
related https://github.com/neuromation/platform-admin/issues/80

neuromation/cli/job.py
```diff
@@ -954,13 +955,21 @@ def format_fail(job: str, reason: Exception) -> str:
         available = await root.client.users.get_acl(
             root.client.username, scheme="storage"
         )
+        to_skip = set()
```
serhiy-storchaka: Would not be simpler to collect a list of valid uris?
```python
storage_uris = []
for perm in available:
    ...
    storage_uris.append(perm.uri)
```

asvetlov: Done

PR #1247, opened 2019-12-13
Support printing multiline strings in TTYPrinter.

serhiy-storchaka:
This will reduce cursor blinking in the progress of copying files.


asvetlov:
Looks good when green

PR #1252, opened 2019-12-17
Fix styling in tables

### LINKED ISSUE #1245: Table does not respect output of invisible characters
Current "table" formatter does align properly on `click.style()` input. For example:

```
Name                              Role
dalazx                                    admin
tvoinarovskyi                             admin
ajuszkowski                               user
mariyadavydova                            user
iadramelk                                 user
zubenkoivan                               user
alias2696                                 user
```

tvoinarovskyi:
Fixes #1245 

@asvetlov @serhiy-storchaka Hey guys, I did not yet do tests for this but want your opinion if we want this type of a fix. The aim was to support any styling in the table although extracting it from a string seems a bit ugly...

Alternatively, we could do something very simple instead, like:
```python
table(..., header_style="bold")
```
Or add a class StyledCell that can wrap cell text and provide style information explicitly.
```python
headers = (StyledCell("Name", bold=True), StyledCell("Role", bold=True))
table(rows=[headers] + other_rows)
```

### DIFF
--- a/neuromation/cli/utils.py
+++ b/neuromation/cli/utils.py
@@ -785,7 +785,6 @@ class StyledTextHelper:
 
         result: List[str] = []
 
-        print(repr(text), repr(wrapped), sm.get_opcodes())
         for op, ti, tj, wi, wj in sm.get_opcodes():
             if op == "equal" or op == "insert":
                 # For those 2 cases we just take the text wrap() generated for us
--- a/neuromation/cli/utils.py
+++ b/neuromation/cli/utils.py
@@ -754,7 +754,6 @@ class StyledTextHelper:
             else:
                 if len(token) >= remaining:
                     result.append(token[:remaining])
-                    remaining = 0
                     break
                 remaining -= len(token)
                 result.append(token)
--- a/neuromation/cli/utils.py
+++ b/neuromation/cli/utils.py
@@ -9,6 +9,7 @@ import time
 from contextlib import suppress
 from datetime import date, timedelta
 from functools import wraps
+from textwrap import wrap
 from typing import (
     Any,
     Awaitable,
@@ -30,6 +31,7 @@ import click
 import humanize
 import pkg_resources
 from click import BadParameter
+from wcwidth import wcswidth
 from yarl import URL
 
 import neuromation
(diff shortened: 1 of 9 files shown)


asvetlov:
I'm not sure that a style per cell is enough.
I may want to emphasize only one word from the cell for example.
A sophisticated processor that calculates the current style for an arbitrary symbol in the string and splits the string in two keeping styles would be better.

serhiy-storchaka:
Currently `table()` does not depend on `click` and I thought that we can just parametrize it by adding a parameter like `width_func` and pass `width_func=lambda s: len(click.unstyle(s))` to handle the styled text. But it seems that we need to handle also truncating, so the table cannot be made style-agnostic.

Note also that `textwrap.wrap()` can break long words on a middle of the control sequence. Or just break a text between control sequences, so the first one will continue to the rest of the line.

tvoinarovskyi:
@asvetlov By itself, it is not hard to calculate style per each character and I think I can manage that quite efficiently. The problem would mostly be in handling _textwrap_ result, as the character may not be at the same offset in the resulting string (break lines are added, spaces removed, etc.). It is not too hard to match it up, just will need some additional development time.

tvoinarovskyi:
So, for now, I will switch to platform-admin and finish this up at the end of the week or so. Will probably do it as you say, with style per character.

asvetlov:
Cool!
Also, please, replace `len(string)` with `wcwidth(string)` for more precise calculations

serhiy-storchaka:
`textwrap` uses `len` and it is not configurable yet. https://bugs.python.org/issue12499

tvoinarovskyi:
@asvetlov Cool was wondering how to process all Unicode characters correctly, did not know that there's such a convenient tool.

tvoinarovskyi:
@asvetlov Well, the good news is that it kind of works. But it is quite ugly... Any ideas on how to make the processor of text more manageable?))

neuromation/cli/utils.py
```diff
@@ -680,3 +683,164 @@ def pager_maybe(
...
+            return list(cls._reapply_styles(text, wrapped))
+        else:
+            return wrap(text, width)
+
+    @classmethod
+    def _reapply_styles(cls, text: str, wrapped: str) -> Iterator[str]:
+        """ Iterate over wrapped text and original one and reapply styles from
+        the original to wrapped
+        """
+        sm = SequenceMatcher(None, text, wrapped)
```
tvoinarovskyi: Found using difflib to be way more readable than trying to write a simpler one myself.

tvoinarovskyi:
@asvetlov @serhiy-storchaka Would be very grateful if you cool take another look after the last rewrite. It's a bit complicated still, but I am satisfied for now :)

neuromation/cli/utils.py
```diff
@@ -680,3 +683,163 @@ def pager_maybe(
...
+
+    # As we often do operations on headers or similar static text it makes sense to
+    # cache
+    @classmethod
+    @lru_cache(255)
+    def unstyle(cls, text: str) -> str:
+        return click.unstyle(text)
+
+    @classmethod
+    @lru_cache(255)
```
asvetlov: Minor: not sure if the minor speedup worth caching of 255 lines

tvoinarovskyi: Yea, mostly those operations are quite cheap, let's remove the cache for now.  My initiative was that unstyle is actually called quite a lot on the code and maybe it makes sense to avoid reprocessing it.

PR #1254, opened 2019-12-17
Add project slug name to neuro project init

### LINKED ISSUE #1230: Make it easier to determine new project directory after neuro project init
Every time I create a new project, my immediate next action is to `cd` into project directory and either open it in my editor or run `make setup`. Things get more complicated if project slug is different from actual name.

My current workaround is to copy-paste slug suggested from `neuro project init` prompt which adds a manual step and complicates writing tutorials.

I see a few possible solutions:

1. Automatically change into project directory, then the name can be obtained using `${PWD##*/}`. Drawback is some users might not expect this side effect from running the cli. `git init` and [Create React App](https://reactjs.org/docs/create-a-new-react-app.html#create-react-app) don't `cd`.

2. Take directory name as an optional argument in `neuro project init`: check if it's a valid slug and default slug and long name to it. `git init` does the same.

I am favoring no. 2 as it can also make `neuro project init` ask less questions. I am using [React scaffolding](https://reactjs.org/docs/create-a-new-react-app.html#create-react-app) as an example.

anayden:
Closes #1230

### DIFF
--- a/CHANGELOG.D/1230.feature
+++ b/CHANGELOG.D/1230.feature
@@ -1,0 +1,1 @@
+`neuro project init` now supports argument to set default value for generated project directory
--- a/README.md
+++ b/README.md
@@ -713,6 +713,7 @@ neuro project init example
 
 Name | Description|
 |----|------------|
+|_\-q, --quiet_|Don't ask any questions, use default values for project setup|
 |_--help_|Show this message and exit.|
 
 
--- a/README.md
+++ b/README.md
@@ -692,7 +692,7 @@ Initialize an empty project.<br/>
 **Usage:**
 
 ```bash
-neuro project init [OPTIONS]
+neuro project init [OPTIONS] [SLUG]
 ```
 
 **Examples:**
@@ -702,6 +702,10 @@ neuro project init [OPTIONS]
 # Initializes a scaffolding for the new project with the recommended project
 # structure (see http://github.com/neuromation/cookiecutter-neuro-project)
 neuro project init
+
+# Initializes a scaffolding for the new project with the recommended project
+# structure and sets default project folder name to "example"
+neuro project init example
 
 ```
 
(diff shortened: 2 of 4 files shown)


serhiy-storchaka:
Would not the `-q` option be confused with the global `-q` option: `neuro -q project init` and  `neuro project init -q`?

anayden:
This option is not necessary for users: you normally wouldn't want to create a project with default cookiecutter settings (especially, project name). It's only needed in tests.

asvetlov:
Good catch!
I suggest renaming the option with `--no-input`, two options with the same name but different meaning is very confusing.
Also, please consider `hidden=True` for the debug-only option.

ayushkovskiy:
Personally, I disagree with disabling input for `neuro project init` -- we introduced this only to facilitate tests, but in fact this breaks the intended logic -- even if `project-slug` is defined, `project-name` and `code directory` will remain default. So let's drop `-q` functionality until it's released.

anayden:
I don't follow. What's wrong with having this option again?

asvetlov:
Regarding `-q` naming: consider `neuro -q project init -q slug`.
The first `-q` has totally different meaning that the second `-q`.
That's why I object. If we absolutely need this option for testing -- let's rename and hide it.
I don't follow `cookiecutter-template` project too tightly for deciding if the option needed at all or not.
@ayushkovskiy and @anayden please make an agreement on it.

ayushkovskiy:
> I don't follow. What's wrong with having this option again?

`neuro project init -q custom-slug` is equivalent to 
```
project_name [Name of the project]: 
project_slug [custom-slug]: 
code_directory [modules]: 
```
so with `-q` there's no way to set up `project_name` and `code_directory`. 
And in case of `neuro project init -q`, `project_slug` will also become default `name-of-the-project`, which might be surprising for a user.

PR #1257, opened 2019-12-18
Add neuro admin generate-cluster-config command

### LINKED ISSUE #1227: Add `neuro admin generate-cluster-config`
This is an interactive command that asks a user a series of questions and, as a result, writes a combined yaml file under the specified path. If the path was omitted, the default value should be used. the resulting path should be commentated regardless.

the questions should be based on the response from 
```
curl -H "Content-Type: application/json" \
     https://dev.neu.ro/api/v1/cloud_providers
```
 https://github.com/neuromation/platform-config#cloud-provider-configuration-options

the resulting file should have the structure that is expected as a payload to the clusters cloud providers route: https://github.com/neuromation/platform-config#configuring-cloud-provider

the node pool configuration and the storage configuration should be generated based on the defaults in the following doc.

see https://docs.google.com/document/d/1muPCUJzJat6ciCfOe9QbSug6skfYEF1XtUarpiUGu4o/edit# for more details.

asvetlov:
Fixes #1227


asvetlov:
Manually tested, generated config files look ok.
The overall UX is, well, come from 70th.

neuromation/cli/admin.py
```diff
@@ -45,12 +49,121 @@ def admin() -> None:
...
+async def generate_aws() -> str:
+    aws_config_file = pathlib.Path(
+        os.environ.get("AWS_CONFIG_FILE", "~/.aws/credentials")
+    )
+    aws_config_file = aws_config_file.expanduser().absolute()
+    parser = configparser.ConfigParser()
+    parser.read(aws_config_file)
+    args = {}
+    args["vpc_id"] = click.prompt("AWS VPC ID")
+    profile = click.prompt("AWS profile name", default="default")
```
dalazx: might be out of scope. should we ask a user whether they want to use the shared credentials file, or rather infer the credentials from the corresponding env vars, or even prompt for them?

also related to the docs @mariyadavydova

neuromation/cli/admin.py
```diff
@@ -45,12 +49,121 @@ def admin() -> None:
...
+  id: standard
+  capacity_tb: 1
+"""
+
+
+async def generate_gcp() -> str:
+    args = {}
+    args["project_name"] = click.prompt("GCP project name")
+    credentials_file = click.prompt(
+        "Google Credentials JSON",
```
dalazx:
```suggestion
        "Service Account Key File",
```

neuromation/cli/admin.py
```diff
@@ -45,12 +49,121 @@ def admin() -> None:
...
+  min_size: 0
+  max_size: 1
+storage:
+  id: generalpurpose_bursting
+"""
+
+
+async def generate_aws() -> str:
+    aws_config_file = pathlib.Path(
+        os.environ.get("AWS_CONFIG_FILE", "~/.aws/credentials")
```
dalazx: the default file path corresponds to another env var `AWS_SHARED_CREDENTIALS_FILE` according to https://docs.aws.amazon.com/cli/latest/topic/config-vars.html

asvetlov: You are right, thanks!

neuromation/cli/admin.py
asvetlov: Thanks for the suggestion.
I've updated the PR to respect more AWS env vars.

PR #1259, opened 2019-12-18
Fix tests and rename the key for project init

### LINKED ISSUE #1258: Fix broken e2e test on Windows
Was introduced by #1254
@anayden, please fix ASAP

anayden:
Closes #1258

### DIFF
--- a/tests/e2e/test_project.py
+++ b/tests/e2e/test_project.py
@@ -10,21 +10,25 @@ from tests.e2e import Helper
 def test_project_init(helper: Helper) -> None:
     path = pathlib.Path(f"./name-of-the-project")
     assert not path.exists()
+    helper.run_cli(["project", "init", "--no-input"])
+    assert path.exists()
+    assert path.is_dir()
     try:
-        helper.run_cli(["project", "init", "-q"])
-        assert path.exists()
-        assert path.is_dir()
-    finally:
         shutil.rmtree(path)
+    except Exception:
+        # ignore errors, the directory may not exist legitimately
+        pass
 
 
 @pytest.mark.e2e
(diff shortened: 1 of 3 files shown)


tests/e2e/test_project.py
```diff
@@ -10,21 +10,25 @@
...
+    assert path.exists()
+    assert path.is_dir()
     try:
-        helper.run_cli(["project", "init", "-q"])
-        assert path.exists()
-        assert path.is_dir()
-    finally:
         shutil.rmtree(path)
+    except Exception:
+        # ignore errors, the directory may not exist legitimately
```
serhiy-storchaka: If the directory may not exist then `assert path.is_dir()` can fail?

tests/e2e/test_project.py
```diff
@@ -11,20 +11,22 @@ def test_project_init(helper: Helper) -> None:
     path = pathlib.Path(f"./name-of-the-project")
     assert not path.exists()
     try:
-        helper.run_cli(["project", "init", "-q"])
+        helper.run_cli(["project", "init", "--no-input"])
         assert path.exists()
         assert path.is_dir()
     finally:
-        shutil.rmtree(path)
+        if path.exists():
```
serhiy-storchaka: It still does not look right. If `path.exists()` can be false then `assert path.exists()` will fail. If remove asserts, we will loss the way to check that `project init` works.

anayden: What I observe now is that `project init` fails with exception on Windows. I'm investigating that first.

ayushkovskiy:
Let's close this PR in favour of https://github.com/neuromation/platform-client-python/pull/1260.

anayden:
Done

Issue #1189, opened 2019-11-22
Increase length of the NAME column in `nuero ps`


ayushkovskiy:
Let's increase the space for the name so that *most* job names are not shifted to the new line. Pesonally, I usually use double-click on the job name to "select the whole job name" and then use it. But when it's shifted to another line, double-click doesn't work.

Currently, the name len is `20` chars out of possible `40`.  In [Template](https://github.com/neuromation/cookiecutter-neuro-project/blob/06aee7e4a384ea1d481b80f738331f97b2de10f6/%7B%7Bcookiecutter.project_slug%7D%7D/Makefile#L18-L22), we prefix all jobs, and the longest prefix is `len("tensorboard-") == 12`. Plus, the project name, which we don't expect to be too short (let's reserve around 25: `len('my-cool-pytorch-project') == 23`).

So here I propose to forbid shifting job name to a new line in `neuro ps` output.

Note: if the line becomes too long, we could shorten other fields and finally remove the Description whom no one use.

PR #1274, opened 2019-12-30
Adjust NAME column of `neuro ps` so that grep by name works.

ayushkovskiy:
Finally closes https://github.com/neuromation/platform-client-python/issues/1189


ayushkovskiy:
@asvetlov please merge if it's OK

PR #1288, opened 2020-01-09
Ps columns

asvetlov:
Add `--format` option to specify `neuro ps` table columns.

The PR is functional; need to add a help topic about format specifiers structure.

### DIFF
--- a/neuromation/cli/parse_utils.py
+++ b/neuromation/cli/parse_utils.py
@@ -77,10 +77,15 @@ COLUMNS = [
 
 COLUMNS_MAP = {column.id: column for column in COLUMNS}
 
-COLUMNS_SPLIT_RE = re.compile(r"(?:\s*,+\s*)|(?:\s+)")
+COLUMNS_RE = re.compile(
+    r"""
+    (?P<col>(\{[^}]+)\})|
+    (?P<sep>\s*,?\s*)|
+    (?P<miss>.)
+    """, re.VERBOSE)
 COLUMN_RE = re.compile(
     r"""
-    \A\{
+    \A
     (?P<id>\w+)
     (?:
       (?:;align=(?P<align>\w+))|
@@ -88,8 +93,8 @@ COLUMN_RE = re.compile(
       (?:;max=(?P<max>\w+))|
       (?:;width=(?P<width>\w+))
     )*
-    (?:;(?P<title>\w+))?
-    \}\Z
+    (?:;(?P<title>.+))?
+    \Z
     """,
     re.VERBOSE,
 )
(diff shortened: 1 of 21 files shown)


neuromation/cli/parse_utils.py
```diff
@@ -46,3 +53,100 @@ def parse_memory(memory: str) -> int:
...
+)
+
+
+def _get(
+    dct: Dict[str, str],
+    name: str,
+    fmt: str,
+    converter: Callable[[str], _T],
+    default: _T,
+) -> Optional[_T]:
```
serhiy-storchaka: According to other type annotations the function never returns None. Either the return annotation should be without `Optional` or `Optional` is needed for `converter` or `default`.

neuromation/cli/parse_utils.py
```diff
@@ -46,3 +53,100 @@ def parse_memory(memory: str) -> int:
...
+    JobColumnInfo("cluster_name", "CLUSTER", Align.LEFT, ColumnWidth(max=15)),
+    JobColumnInfo("description", "DESCRIPTION", Align.LEFT, ColumnWidth(max=50)),
+    JobColumnInfo("command", "COMMAND", Align.LEFT, ColumnWidth(max=100)),
+]
+
+COLUMNS_MAP = {column.id: column for column in COLUMNS}
+
+COLUMNS_RE = re.compile(
+    r"""
+    (?P<col>(\{[^}]+)\})|
```
serhiy-storchaka: Inner `(` and `)` are not needed.

Besides, it may be more convenient to use `\{(?P<col>[^}]+)\}|`, so you will not need to strip the first and the last characters by `[1:-1]`.

neuromation/cli/parse_utils.py
```diff
@@ -46,3 +53,100 @@ def parse_memory(memory: str) -> int:
...
+    )*
+    (?:;(?P<title>.+))?
+    \Z
+    """,
+    re.VERBOSE,
+)
+
+
+def _get(
+    dct: Dict[str, str],
```
serhiy-storchaka:
```suggestion
    dct: Dict[str, Optional[str]],
```
An unmatched group is None by default (can be overridden by passing the `default` argument to `groupdict()`).

neuromation/cli/parse_utils.py
```diff
@@ -46,3 +53,100 @@ def parse_memory(memory: str) -> int:
...
+    name: str,
+    fmt: str,
```
serhiy-storchaka: `fmt` is not used.

neuromation/cli/parse_utils.py
asvetlov: Let me change the type of `default`

neuromation/cli/parse_utils.py
asvetlov: Fixed, I wanted to use `fmt` for the error text formatting.

neuromation/cli/parse_utils.py
asvetlov: done

neuromation/cli/parse_utils.py
```diff
@@ -46,3 +53,100 @@ def parse_memory(memory: str) -> int:
...
+            continue
+        elif m1.lastgroup == "miss":
+            raise ValueError(f"Invalid format {fmt!r}")
+        column = m1.group()[1:-1]
+        m2 = COLUMN_RE.fullmatch(column)
+        if m2 is None:
+            raise ValueError(f"Invalid format {fmt!r}")
+        groups = m2.groupdict()
+        id = groups["id"]
+        if id not in COLUMNS_MAP:
```
serhiy-storchaka: I would make column names case-insensitive and allow abbreviations:
```python
id = id.lower()
default = COLUMNS_MAP.get(id)
if default is None:
    for name in COLUMNS_MAP:
        if name.startswith(id):
            if default is not None:
                raise ValueError(f"Ambiguous column {id!r} in format {fmt!r}")
            default = COLUMNS_MAP[id]
    if default is None:
        raise ValueError(f"Unknown column {id!r} in format {fmt!r}")
```
But this can be added in a separate PR.

asvetlov: Excellent idea!
Implemented and tested.

asvetlov:
e2e tests are broken by `neuro top` command.
The failure is not related to the PR, I prefer to fix it in separate pull request.

PR #1289, opened 2020-01-09
extend admin get-clusters command result with additional information

### LINKED ISSUE #1287: neuro admin get-clusters should use the clusters API
since https://github.com/neuromation/platform-config/issues/273 is merged, we want to improve the way we communicate the state of the clusters in `neuro admin get-clusters`:

we need to:
- [ ] patch `_Admin.list_clusters` to use `/api/v1/clusters` instead of the admin API: https://github.com/neuromation/platform-client-python/blob/df7e6c76661a4dada22158e14d9fecce4bb529d2/neuromation/api/admin.py#L36-L45
- [ ] expand `_Cluster` and add the `status` attribute, as well as the infra info such as `node_pools` and `storage` there;
- [ ] patch the corresponding formatter to output all the new information added above ^.

zubenkoivan:
closes #1287

### DIFF
--- a/neuromation/cli/formatters/admin.py
+++ b/neuromation/cli/formatters/admin.py
@@ -26,7 +26,7 @@ class ClustersFormatter:
         out = []
         for cluster in clusters:
             out.append(style("Name: ", bold=True) + cluster.name)
-            out.append(style("Status: ", bold=True) + cluster.status)
+            out.append(style("Status: ", bold=True) + cluster.status.capitalize())
             if cluster.cloud_provider:
                 cloud_provider = cluster.cloud_provider
                 out.append(style("Cloud: ", bold=True) + cloud_provider.type)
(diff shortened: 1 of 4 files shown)


neuromation/cli/formatters/admin.py
```diff
@@ -20,8 +23,108 @@ def __call__(self, clusters_users: Iterable[_ClusterUser]) -> List[str]:
...
+
+def _no() -> str:
+    return "No" if sys.platform == "win32" else "✖︎"
+
+
+def _has_tpu(node_pools: Iterable[_NodePool]) -> bool:
+    has_tpu = False
+    for node_pool in node_pools:
+        if node_pool.is_tpu_enabled:
+            has_tpu = True
```
serhiy-storchaka: Why not just `return True`?

serhiy-storchaka:
:+1:

zubenkoivan:
@serhiy-storchaka thanks!

PR #1290, opened 2020-01-10
Simple columns

asvetlov:
Omit curly brackets if only column ids are omitted.


neuromation/cli/parse_utils.py
```diff
@@ -82,6 +82,7 @@ class JobColumnInfo:
 
 COLUMNS_RE = re.compile(
     r"""
+    (?P<id>\w+)|
     (?:\{(?P<col>[^}]+)\})|
     (?P<sep>\s*(?:,\s*)?)|
```
serhiy-storchaka: I think this should fix a bug on 3.6:
```suggestion
    (?P<sep>\s*(?:,\s*|\s))|
```
`sep` is never an empty string, it contains a comma or at least one space.

asvetlov: Fixed by 8dad20b
Thanks!

PR #1296, opened 2020-01-16
Allow blank lines and comments in --env-file.

### LINKED ISSUE #1208: Allow empty lines in `neuro run --env-file`
I want the env-file to allow empty lines because it's a real use-case to have empty lines:
1. separator
2. lack of trailing emptyline is hard to maintain in an automated workflow.

```
[ay@archlinux ~]$ cat env
key=val
val2

[ay@archlinux ~]$ neuro run -s cpu-small -n test-env --env-file=env ubuntu "true"
ERROR: Illegal argument(s) ({'container': DataError({'env': DataError({'': DataError({'key': DataError(blank value is not allowed)})})})})
[ay@archlinux ~]$ vim env
```

serhiy-storchaka:
Closes #1208.

### DIFF
--- a/neuromation/cli/job.py
+++ b/neuromation/cli/job.py
@@ -7,7 +7,7 @@ import sys
 import textwrap
 import uuid
 import webbrowser
-from typing import Dict, List, Optional, Sequence, Set, Tuple
+from typing import Dict, Iterator, List, Optional, Sequence, Set, Tuple
 
 import async_timeout
 import click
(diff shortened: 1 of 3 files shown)


neuromation/cli/job.py
```diff
@@ -80,10 +80,18 @@ def _get_neuro_mountpoint(username: str) -> str:
     return f"{ROOT_MOUNTPOINT}/{username}"
 
 
+def _read_lines(env_file: str) -> Iterator[str]:
+    with open(env_file, "r") as ef:
```
asvetlov: Should the file be opened with explicit "utf8" mode?

neuromation/cli/job.py
```diff
@@ -80,10 +80,18 @@ def _get_neuro_mountpoint(username: str) -> str:
...
+        lines = ef.read().splitlines()
```
asvetlov: Minor: `ef.readline()` can be used here.
I don't care too much though because the file with env vars is expected to be short, the memory footprint is not a problem here.

PR #1298, opened 2020-01-16
Use Sqlite for saving the configuration file

asvetlov:


asvetlov:
I've just moved existing yaml as is into sqlite db column.
Let's decouple this big yaml into separate tables and fields in future pull requests.

serhiy-storchaka:
What is an advantage of using Sqlite over plain file?

Currently, the configuration file contains two kinds of data: data, generated automatically or received from the server (URLs, tokens, expiration dates, cluster parameters, etc) and data which should be added by the user -- aliases, default filters. It is difficult to edit an Sqlite database to set default filters.

asvetlov:
> Currently, the configuration file contains two kinds of data: data, generated automatically or received from the server (URLs, tokens, expiration dates, cluster parameters, etc) and data which should be added by the user -- aliases, default filters.

Exactly!

The PR is for using Sqlite for the first category (received from the server or autogenerated data), TOML is for user-provided configuration variables.

serhiy-storchaka:
But why Sqlite?

neuromation/api/config.py
```diff
@@ -449,3 +465,35 @@ def _load_file(filename: Path) -> Mapping[str, Any]:
     config = toml.load(filename)
     _validate_user_config(config, filename)
     return config
+
+
+def _load_schema(db: sqlite3.Connection) -> Dict[str, str]:
+    cur = db.cursor()
+    schema = {}
+    cur.execute("SELECT type, name, sql from sqlite_master")
+    for row in cur:
```
serhiy-storchaka: You can also unpack a row:
```suggestion
    for type, name, sql in cur:
```

neuromation/api/config.py
```diff
@@ -266,18 +272,28 @@ def _save(cls, config: _Config, path: Path) -> None:
...
+            _init_db_maybe(db)
+
+            cur = db.cursor()
+            content = yaml.safe_dump(payload, default_flow_style=False)
+            cur.execute("DELETE FROM main")
+            cur.execute(
+                """
+                INSERT INTO main (content, timestamp)
+                VALUES (?, ?)""",
+                (content, time.time()),
```
serhiy-storchaka: `time.time()` is float, but the timestamp column is INTEGER. I suppose this will cause a warning in 3.8 and error in 3.10.

Where the timestamp is used?

asvetlov: Nowhere. I wanted to save the last update time. 
The second-round precision is good enough, but I can change the type to REAL

PR #1305, opened 2020-01-21
Send the usage statistics to Google Analytics

### LINKED ISSUE #1286: Add Google Measurement Protocol into CLI
To improve user experience we would like to collect anonymous usage data using [Google Measurement Protocol](https://developers.google.com/analytics/devguides/collection/protocol/v1).

After each CLI run we should submit an [event into Google Analytics](https://developers.google.com/analytics/devguides/collection/protocol/v1/devguide#event) with category `CLI`, action `Run`, and label being a JSON capturing command line arguments and exit code. Tracking id: `UA-106571369-3`

Anonymous User ID that is passed with the HTTP call needs to be auto-generated and persisted locally.

Usage reporting should be turned on by default, however, users should be able to opt-out from sending usage data by running `neuro config opt-out`.

Upon first run of the CLI we should display the message saying: `In order to improve user experience, we are anonymously tracking usage data. You can opt-out by running "neuro confing opt-out"`. The message should be displayed once after first installation of `neuro`.

asvetlov:
Fix #1286

### DIFF
--- a/neuromation/cli/main.py
+++ b/neuromation/cli/main.py
@@ -198,7 +198,7 @@ def print_options(
     ),
 )
 @option(
-    "--skip-stats/--no-skip-token",
+    "--skip-stats/--no-skip-stats",
     is_flag=True,
     default=False,
     help=(
--- a/neuromation/cli/main.py
+++ b/neuromation/cli/main.py
@@ -197,6 +197,17 @@ def print_options(
         "together with option '--trace'. On by default."
     ),
 )
+@option(
+    "--skip-stats/--no-skip-token",
+    is_flag=True,
+    default=False,
+    help=(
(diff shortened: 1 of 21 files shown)


asvetlov:
The PR is tested manually, I'm working on adding some unit tests.

neuromation/cli/main.py
```diff
@@ -196,6 +197,17 @@ def print_options(
         "together with option '--trace'. On by default."
     ),
 )
+@option(
+    "--skip-stats/--no-skip-token",
```
anayden: Looks like a typo: should be `--no-skip-stats`

asvetlov: You are right, thanks!

asvetlov:
Tests are done I think

PR #1320, opened 2020-02-05
Custom command aliases

asvetlov:


neuromation/cli/alias.py
```diff
@@ -0,0 +1,324 @@
...
+            else:  # pragma: no cover
+                # Unreachable code, _validate_exec()
+                # makes sure that all param names are handled
+                raise ConfigError(f'Unknown parameter {name} in "{cmd}"')
+            val = ctx.params[name]
+            replaces[match] = " ".join(
+                shlex.quote(arg) for arg in _process_param(param, val)
+            )
+
+        for name, val in replaces.items():
```
serhiy-storchaka: What if the same name is occurred several times in `cmd`?

PR #1325, opened 2020-02-11
Add job run-time limit flag to `neuro run`

ayushkovskiy:
Support run-time limit enforcement functionality: https://github.com/neuromation/platform-api/pull/1044


neuromation/api/utils.py
```diff
@@ -108,3 +109,7 @@ def retries(
         sleeptime += 0.1
         attempts -= 1
         yield retry()
+
+
+def _delta_total_minutes(delta: timedelta) -> int:
+    return int(delta.total_seconds() // 60)
```
ayushkovskiy: should this be in `neuromation/api/jobs.py`, where it is used?

neuromation/cli/job.py
```diff
@@ -739,6 +751,14 @@ def format_fail(job: str, reason: Exception) -> str:
...
     secure=True,
 )
+@option(
+    "-l",
+    "--limit",
+    metavar="MINUTES",
+    type=int,
+    help="Optional job's run-time limit in minutes (set '0' to disable)",
+    default=None,
+)
```
ayushkovskiy: I am not sure if the names are good, let's change in case of any objection:
1. `-l/--limit` or `--timeout`? or `--max-run-time`
2. is `0` good to disable the feature, or should we have a separate flag `--no-limit`?
3. is limit in minutes good? note, the server expects the value in minutes, so there's no reason to have user-faced seconds.

serhiy-storchaka:
You can use the [timeout](http://man7.org/linux/man-pages/man1/timeout.1.html) command.

ayushkovskiy:
> You can use the timeout command.

@serhiy-storchaka sorry I didn't understand you: for what?

serhiy-storchaka:
I meant that the user can use the `timeout` command instead of `neuro` option to limit the job runtime.
```
$ neuro run timeout 1h job-command
```
If add such feature in `neuro`, I think it should time suffixes `m` for minutes and `h` for hours as in `admin add-user-quota`. It is also compatible with `timeout`.

serhiy-storchaka:
I think `--limit` is not good name. We can add other limits in future.

neuromation/api/utils.py
asvetlov: good idea

neuromation/cli/job.py
asvetlov: `--no-limit` is not needed

neuromation/api/config.py
```diff
@@ -133,6 +133,7 @@ class _Config:
     version: str
     cluster_name: str
     clusters: Mapping[str, Cluster]
+    default_run_time_limit: timedelta
```
asvetlov: Why `default_run_time_limit` is a part of config?
It is explicitly specified by `jobs.run()` argument.
If you want to provide user-default value for the limit -- see `neuromation/cli/jobs:calc_filters` for how to do it properly.

ayushkovskiy:
@asvetlov, a few questions if you don't mind:

1. Any objections from using [pytimeparse](https://pypi.org/project/pytimeparse/)? It provides the parser for free format time-delta definition like `2d 5h 34m` or `2 days, 5 hours, 34 minutes` or `5hr34m56s`. However, it's officially supported only up to Pyhton 3.6. If we refuse to use it, I'll write similar but simplified regex-based parser to parse strings like `2d 5h 34m`, not a problem.
2. If we change the name of the flag to `neuro run --timeout`, is that a problem that `neuro exec --timeout` already exists, but `--timeout` there has slightly different meaning?
3. How can we (and should we) write the default config value to the user's yml-config? I see method `calc_columns` addresses the default value `COLUMNS` if no value found in config, so here we repeat this algorithm for method `calc_default_job_timeout`. However, I thought it might be useful if we write this default value to the yml-config, and just let the user modify it -- it's easier.

asvetlov:
> @asvetlov, a few questions if you don't mind:
> 
> 1. Any objections from using [pytimeparse](https://pypi.org/project/pytimeparse/)? It provides the parser for free format time-delta definition like `2d 5h 34m` or `2 days, 5 hours, 34 minutes` or `5hr34m56s`. However, it's officially supported only up to Pyhton 3.6. If we refuse to use it, I'll write similar but simplified regex-based parser to parse strings like `2d 5h 34m`, not a problem.

A free format is not good for usage in CLI parameters: `2d 5h 34m` should be enclosed in brackets, `2 days, 5 hours, 34 minutes` is even worse because of its length.

> 2. If we change the name of the flag to `neuro run --timeout`, is that a problem that `neuro exec --timeout` already exists, but `--timeout` there has slightly different meaning?

Please use `--life-span`, the term is very well known, short enough and unambiguous.

> 3. How can we (and should we) write the default config value to the user's yml-config? I see method `calc_columns` addresses the default value `COLUMNS` if no value found in config, so here we repeat this algorithm for method `calc_default_job_timeout`. However, I thought it might be useful if we write this default value to the yml-config, and just let the user modify it -- it's easier.

The philosophy of user-provided configuration files is that they are never modified by the application itself. Even a command for user config file changing is redundant -- every software engineer knows how to use a text editor. Please use the hard-coded default value if absent.

P.S. I wonder why do you use `yml-config` term for `toml` file format?

ayushkovskiy:
Re-worked this PR considering the comments.
1. Not a big deal not to use a side library, the hand-written parser turned out to be small and supporting more concise format: `1d 2h 3m 4s` or `1d2h3m4s` or `1d 3m` or `1d` or `0`  (with or without spaces, where some parts may be missing, or `0` as an edge-case since `0s` is the same as `0m`).
2. I never heard of `--life-span` in the sense of timeout, but it does not matter for me, I did the `--life-span`.
3. I would think of a user-config as of a normal Unix config (a `.conf` file somewhere inside `/etc/`), which contains all sections with defaults in the right format but commented out. But I agree that, comparing to Unix utilities, `neuro` is evolving fastly, and the format of the user-config is changed fastly either. So this is apparently the part of documentation, to teach user the format. 
BTW, note the format of the new section:
```
[job.default-life-span]
days = 1
hours = 2
minutes = 3
```

> P.S. I wonder why do you use yml-config term for toml file format?
my bad, a typo-like mistake.

ayushkovskiy:
New question: What should be the short option for `--life-time`? Should it be at all?

neuromation/cli/job.py
```diff
@@ -1127,3 +1176,54 @@ def calc_statuses(status: Sequence[str], all: bool) -> Set[JobStatus]:
...
+    result = (
+        _parse_timedelta(value)
+        if value is not None
+        else await calc_default_life_span(client)
+    )
+    seconds = result.total_seconds()
+    if seconds == 0:
+        return timedelta.max
+    elif seconds < 0:
+        raise click.UsageError(f"Job life span must be non-negative, got: '{result}'")
```
ayushkovskiy: technically, our input parser does not allow negative values, so they can come only from the config. I understand, it's better to check it during user config validation, but it's quicker to do it here, after parsing.
1. Should I add more validation code to `_validate_user_config()`, similarly as it is with `_validate_alias()`?
2. Why don't we use `trafaret`? Doesn't it have enough expressive power to validate aliases, does it?

asvetlov:
> New question: What should be the short option for `--life-time`? Should it be at all?

If you are in doubt -- don't use short name, that's fine.

neuromation/api/config.py
```diff
@@ -455,7 +458,20 @@ def _validate_user_config(
     # additional API for describing new supported config sections, keys and values.
     # Right now this functionality is skipped for the sake of simplicity.
     _check_sections(config, {"alias", "job", "storage"}, filename)
-    _check_section(config, "job", {"ps-format": str}, filename)
+    _check_section(
+        config,
+        "job",
+        {
+            "ps-format": str,
+            "default-life-span": {
```
asvetlov: Please use `"life-span": str` here, where `str` is exactly the same string as passed to `neuro run --life-span TIMEDELTA`.

neuromation/api/jobs.py
```diff
@@ -137,6 +137,7 @@ def __init__(self, core: _Core, config: Config, parse: Parser) -> None:
         description: Optional[str] = None,
         is_preemptible: bool = False,
         schedule_timeout: Optional[float] = None,
+        life_span: timedelta = timedelta.max,
```
asvetlov: Please use `Optional[float]` type here, like for `schedule_timeout`. `None` means *no lifespan limit`.
`float` means *time in seconds*, not *amount of minutes*.

neuromation/cli/job.py
```diff
@@ -278,6 +298,7 @@ def job() -> None:
     volume: Sequence[str],
     env: Sequence[str],
     env_file: Optional[str],
+    life_span: str,
```
asvetlov:
```suggestion
    life_span: Optional[str],
```

neuromation/cli/job.py
```diff
@@ -243,6 +250,19 @@ def job() -> None:
...
+    "--life-span",
+    type=str,
+    metavar="TIMEDELTA",
+    help=(
+        "Optional job run-time limit in the format '1d 2h 3m 4s' "
+        "with or without spaces, some parts may be missing. "
+        "Set '0' to disable. "
+        "(default value can be changed in the user config)"
+    ),
+    default=DEFAULT_JOB_LIFE_SPAN,
```
asvetlov: Drop `default`. Use `None` as a marker for a case when life-span is not explicitly provided by a user.
Use either config or application-level default then.

neuromation/cli/job.py
```diff
@@ -889,6 +927,13 @@ def format_fail(job: str, reason: Exception) -> str:
     if not wait_start:
         detach = True
 
+    job_life_span = await calc_life_span(root.client, life_span)
+    if not root.quiet:
```
asvetlov: I rather prefer to see this printout in explicit verbose mode only. In both normal and quiet modes the message should be absent.

neuromation/cli/job.py
asvetlov: > technically, our input parser does not allow negative values, so they can come only from the config. I understand, it's better to check it during user config validation, but it's quicker to do it here, after parsing.
> 
> 1. Should I add more validation code to `_validate_user_config()`, similarly as it is with `_validate_alias()`?

Please keep as is until we implement something like `neuro config validate` command.

> 2. Why don't we use `trafaret`? Doesn't it have enough expressive power to validate aliases, does it?

Trafaret provides machine-readable error messages which should be translated to human-readable texts before printing. It makes `trafaret` virtually useless for our needs.

neuromation/cli/job.py
ayushkovskiy: Isn't the information that your job will die automatically after certain (in some cases, implicit) amount of time a more important information rather than, say, Shortcuts? I think, if the job has a timeout, it should be printed in non-quiet mode. Other question is, perhaps, a better place for it is `formatters/jobs.py`?

neuromation/api/config.py
ayushkovskiy: ok

neuromation/api/jobs.py
ayushkovskiy: no difference, will do.

ayushkovskiy: (I thought `timedelta` is the most convenient way to store time intervals, but ok will do in seconds, `float`)

neuromation/cli/job.py

neuromation/cli/job.py
asvetlov: There is a lot of important information about the job. We need a balance between insufficient and too verbose outputs.
Moving reports about remaining time is a very cool idea.
I see the REST API inconsistency here though: the deadline timestamp is not returned in `JobDescription` data structure. Seems like the server doesn't file this info, that's sad.  In my mind, it should be an absolute UNIX timestamp; relative times are not reliable. The error is one less significant digit due to network latency, 1 minute in your case.
It raises another question: should the absolute timestamp be used as input in REST API?  Passing a relative time in Python API is ok.

neuromation/cli/job.py
```diff
@@ -1127,3 +1169,48 @@ def calc_statuses(status: Sequence[str], all: bool) -> Set[JobStatus]:
...
+        return timedelta(0)
+    match = REGEX_JOB_LIFE_SPAN.search(value)
+    if match is None:
+        raise click.UsageError(
+            f"{err}: Should be like '1d 2h 3m 4s' with or without spaces, "
+            "some parts may be missing"
+        )
+    return timedelta(
+        days=int(match.group("d") or 0),
+        hours=int(match.group("h") or 0),
```
asvetlov: Is "1d35h120m" valid?

ayushkovskiy: yes, I think it should be valid

ayushkovskiy: The order matters, so `1h1d` will be invalid. But I think it's OK

neuromation/cli/job.py
```diff
@@ -75,6 +77,11 @@
 NEUROMATION_HOME_ENV_VAR = "NEUROMATION_HOME"
 RESERVED_ENV_VARS = {NEUROMATION_ROOT_ENV_VAR, NEUROMATION_HOME_ENV_VAR}
 
+DEFAULT_JOB_LIFE_SPAN = "1d"
+REGEX_JOB_LIFE_SPAN = re.compile(
+    r"^((?P<d>\d+)d)?\s*((?P<h>\d+)h)?\s*((?P<m>\d+)m)?\s*((?P<s>\d+)s)?$"
```
asvetlov: Please drop the optional spaces.
I know you can support this mode easily but the case doesn't make sense: wrapping an option value in quotes is cumbersome.
I don't see the motivation for support of spaces except for *just we can do it*.

neuromation/cli/job.py
```diff
@@ -1127,3 +1169,48 @@ def calc_statuses(status: Sequence[str], all: bool) -> Set[JobStatus]:
...
+    assert seconds > 0, f"life-span cannot be negative, got: {seconds}"
+    return seconds
+
+
+async def calc_default_life_span(client: Client) -> timedelta:
+    config = await client.config.get_user_config()
+    section = config.get("job")
+    life_span = DEFAULT_JOB_LIFE_SPAN
+    if section is not None:
+        value = section.get("life-span")
```
asvetlov: Please document this parameter in `neuro help user-config`

neuromation/api/jobs.py
```diff
@@ -137,6 +137,7 @@ def __init__(self, core: _Core, config: Config, parse: Parser) -> None:
         description: Optional[str] = None,
         is_preemptible: bool = False,
         schedule_timeout: Optional[float] = None,
+        life_span: Optional[float] = None,
```
asvetlov: Please document this argument in https://neuromation-sdk.readthedocs.io/en/latest/jobs_reference.html#neuromation.api.Jobs.run

neuromation/cli/job.py
ayushkovskiy: wow, didn't know about this command. OK

neuromation/api/jobs.py
ayushkovskiy: sure!

neuromation/cli/job.py

docs/jobs_reference.rst
```diff
@@ -134,6 +135,9 @@ Jobs
                                      cannot be scheduled because the lack of computation
                                      cluster resources (memory, CPU/GPU etc).
 
+      :param float life_span: job run-time limit in the format '1d2h3m4s'
```
asvetlov: The `'1d2h3m4s'` is invalid float.
What is `None` for? I assume it is for server default, please correct me if I'm wrong.

docs/jobs_reference.rst
```diff
@@ -134,6 +135,9 @@ Jobs
...
+                              (some parts may be missing). Pass '0' to disable.
```
asvetlov: `'0'` or `0`?

docs/jobs_reference.rst
ayushkovskiy: oops

docs/jobs_reference.rst
ayushkovskiy: right.

ayushkovskiy: None

docs/jobs_reference.rst
ayushkovskiy: `None` to disable. I'd actually leave the default API value for `1d`, not for `None` as it is now.

asvetlov: Why not enforce the default on server side?
It is more reliable.
I feel potential problems if we want to change 1day to another default eventually.

ayushkovskiy: Long time ago, when I was implementing this feature [on the backend](https://github.com/neuromation/platform-api/pull/1044), my very first version enforced default server-side as you suggest. However, after several review we agreed to pass default value by the client so that:
1. no problems with DB migration, or if we decide not to write the default value to the database, then no implicit logic;
2. much less code
3. not over-complicated protocol (`0` means `limit=0`)
4. easier to change the default value per-user: no need to put it to the database.

neuromation/cli/job.py
ayushkovskiy: Why not to use [`extract_error`](https://trafaret.readthedocs.io/en/latest/api/trafaret.html#trafaret.extract_error)? It might be easier to use a package that was designed to write custom validators, rather than to implement our own from scratch.

neuromation/cli/job.py
```diff
@@ -1127,3 +1167,47 @@ def calc_statuses(status: Sequence[str], all: bool) -> Set[JobStatus]:
...
+        else await calc_default_life_span(client)
+    )
+    seconds = delta.total_seconds()
+    if seconds == 0:
+        return None
+    assert seconds > 0, f"life-span cannot be negative, got: {seconds}"
+    return seconds
+
+
+async def calc_default_life_span(client: Client) -> timedelta:
```
asvetlov: Inline the function in `calc_life_span()` please.

neuromation/cli/job.py
```diff
@@ -1127,3 +1167,47 @@ def calc_statuses(status: Sequence[str], all: bool) -> Set[JobStatus]:
...
+async def calc_life_span(client: Client, value: Optional[str]) -> Optional[float]:
+    delta = (
+        _parse_timedelta(value)
+        if value is not None
+        else await calc_default_life_span(client)
+    )
+    seconds = delta.total_seconds()
+    if seconds == 0:
+        return None
+    assert seconds > 0, f"life-span cannot be negative, got: {seconds}"
```
asvetlov: Please raise `ValueError` instead of `AssertionError` here.

neuromation/cli/job.py
ayushkovskiy: Why? it's harder to test

neuromation/cli/job.py
ayushkovskiy: It's impossible situation, and we need this check only in case in order to protect ourselves from sending negative values to the server one day if we change something in future (for example, we are going to change toml parsers). So I cannot test this check properly. Here's what we can do:
1. leave assertion
2. change to ValueError but still keep it untested
3. remove the check completely

neuromation/cli/job.py
asvetlov: The function is never used alone, it is always called from `calc_life_span()`.
You can test `calc_life_span(client, None)` instead of `calc_default_life_span(client)`, should prodice the same result.

neuromation/cli/job.py
asvetlov: I read `assert expr, msg` as something that you want to expose to a user.
If the situation is impossible I expect bare `assert expr` at least. Like for the sake of `mypy` correctness, we have `assert var is not None` very often without a custom provided error message.

ayushkovskiy: I was always thinking that `assert expr` is a bad practice since if anything ever goes wrong, the user will not get any idea on the reason, just `AssertionError`.

ayushkovskiy: changed to bare `assert expr`

neuromation/cli/job.py
ayushkovskiy: the matter of taste. changed.

PR #1343, opened 2020-02-19
Fix failed e2e test for --pass-config

asvetlov:


tests/e2e/assets/neuromation-client/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.6-slim
+FROM python:3.7-slim
```
serhiy-storchaka: Is it required? 3.6 is the minimal requirement for the client.

tests/e2e/test_e2e_jobs.py
```diff
@@ -669,7 +670,10 @@ def test_job_run(helper: Helper) -> None:
 ) -> AsyncIterator[str]:
     image = await generate_image(docker)
     yield image
-    await docker.images.delete(image, force=True)
+    try:
```
serhiy-storchaka: While you are here, do you mind to add `try`/`finally` around `yield image`? I was going to do this in a separate PR.

tests/e2e/assets/neuromation-client/Dockerfile
asvetlov: Technically 3.6 works as well. For 3.7 the installation is slightly faster because a couple of packages are skipped.

tests/e2e/test_e2e_jobs.py
asvetlov: I've double-checked recently: `pytest` executes fixutre's tear-down part even if the test raises an exception (doesn't matter is it `AssertionError` or another exception type).
No need to wrap `yield` into `try/finally` block, fixture executor behaves slightly different than `contextlib.[async]contextmanager`.

PR #1353, opened 2020-02-26
Fix test_grant_complete_lifecycle

serhiy-storchaka:
(if it is broken)


serhiy-storchaka:
Well, it is not broken.

PR #1354, opened 2020-02-26
Replacement PR for #1349 (neuro images ls -l support) 

olegstepanov:
Had to recreate PR because Azure failed to run the checks

### DIFF
--- a/neuromation/cli/formatters/images.py
+++ b/neuromation/cli/formatters/images.py
@@ -3,6 +3,8 @@ from typing import Dict, Iterable
 
 import click
 
+from neuromation.cli.formatters.ftable import table
+
 from neuromation.api import (
     AbstractDockerImageProgress,
     ImageProgressPull,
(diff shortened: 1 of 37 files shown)


tests/cli/test_ftable.py
```diff
@@ -58,6 +58,10 @@ def test_multiline(self) -> None:
 
 
 class TestTable:
+    def test_empty(self) -> None:
+        result = list(table(()))
+        assert len(result) == 0
```
serhiy-storchaka:
```suggestion
        assert result == []
```

olegstepanov: Wait, table returns an Iterator[str], is it correct to check that it's an empty list?

serhiy-storchaka: But `result` is a list.

neuromation/cli/job.py
```diff
@@ -545,6 +545,7 @@ def job() -> None:
     """
 
     format = await calc_columns(root.client, format)
+
```
asvetlov:
```suggestion
```

tests/cli/test_ftable.py
olegstepanov: Oops, sorry, missed that

neuromation/cli/formatters/images.py
```diff
@@ -133,3 +135,20 @@ def commit_finished(self, data: ImageCommitFinished) -> None:
...
+
+
+class ShortImagesFormatter(BaseImagesFormatter):
+    def __call__(self, images: Iterable[RemoteImage]) -> Iterable[str]:
+        return (str(image) for image in images)
+
+
+class LongImagesFormatter(BaseImagesFormatter):
+    def __call__(self, images: Iterable[RemoteImage]) -> Iterable[str]:
+        rows = [[str(image), image.as_docker_url] for image in images]
```
asvetlov: The last neat: please add the table header, like for `neuro ps` command.

olegstepanov: What would be good headers? 'IMAGE URL', 'DOCKER URL'?

asvetlov: I guess `Neuro URL` and `Docker URL` (both capitalized) -- to reflect the fact that the first value should be used with `neuro`, the second is for `docker`.

PR #1361, opened 2020-03-02
Steal config files on 'job --pass-config'

asvetlov:


asvetlov:
@dalazx please note, env var and config location are changed for new `--pass-config` implementation.
The change is backward compatible, both old and new CLI can work all together.
The only thing that can be affected is web-shell.

PR #1373, opened 2020-03-05
Resolve only active jobs if needed

asvetlov:


neuromation/cli/job.py
```diff
@@ -392,7 +392,9 @@ def job() -> None:
     neuro exec --no-tty my-job ls -l
     """
     real_cmd = _parse_cmd(cmd)
-    id = await resolve_job(job, client=root.client)
+    id = await resolve_job(
+        job, client=root.client, status={JobStatus.PENDING, JobStatus.RUNNING}
```
ayushkovskiy: neuro exec will fail for status=PENDING, perhaps this should be only "RUNNING"

asvetlov:
Let's wait for the server-side fix, it seems the fix is ready for landing.

PR #1378, opened 2020-03-06
Skip searching by job name if it matches the job id pattern.

serhiy-storchaka:
As suggested by @ayushkovskiy.

It will reduce execution time of every job related command by few seconds.

It is a temporary solution which gives immediate effect until we implement resolving job names at backend.


asvetlov:
The PR looks good, let's merge it

serhiy-storchaka:
I ran tests several times with and without this optimization. Is reduces the run time of e2e-jobs from 14 to 11 minutes (20%). It does not affect the run time of e2e-some (5 minutes).

PR #1391, opened 2020-03-11
[WIP] Object storage

tvoinarovskyi:


tvoinarovskyi:
@asvetlov Still no tests, but could you look through.

neuromation/cli/main.py
```diff
@@ -441,6 +452,8 @@ def help(ctx: click.Context, command: Sequence[str]) -> None:
 cli.add_command(share.acl)
 
 cli.add_command(DeprecatedGroup(storage.storage, name="store", hidden=True))
+if os.environ.get("_NEURO_OBJ_STORAGE"):
```
tvoinarovskyi: Depending on how we merge/release the functionality I kind of did this to hide it during development.

neuromation/cli/object_storage.py
```diff
@@ -0,0 +1,367 @@
...
+    ),
+)
+@option(
+    "-p/-P",
+    "--progress/--no-progress",
+    is_flag=True,
+    default=True,
+    help="Show progress, on by default.",
+)
+async def cp(
```
tvoinarovskyi: Removing functionality actually seemed harder for some cases, like not filtering was hard cause of how all helper functions reuse them. Only removed `update` functionality, as it caused some pain in development.

neuromation/cli/object_storage.py
```diff
@@ -0,0 +1,367 @@
...
+            log.error(f"cannot copy {src} to {dst}: {error}")
+            errors = True
+
+        progress_obj.end()
+
+    if errors:
+        sys.exit(EX_OSFILE)
+
+
+async def _is_dir(root: Root, uri: URL) -> bool:
```
tvoinarovskyi: Those 2 helper functions are behind a validation layer, could be refactored to be usable on both storages instead of a copy.

neuromation/api/object_storage.py
```diff
@@ -0,0 +1,401 @@
...
+                progress.enter(item)
+            elif isinstance(item, StorageProgressLeaveDir):
+                progress.leave(item)
+            elif isinstance(item, StorageProgressFail):
+                progress.fail(item)
+
+    await task
+
+
+async def calc_md5(path: Path) -> str:
```
tvoinarovskyi: Note the MD5 checksum. It's quite different and in Object Storages is optional, but I did not expose an option yet. Do we need it? It's quite fast, but can be intensive for large files (8GB file computes ~16s on my PC. That's 16s without output for user)

neuromation/api/object_storage.py
```diff
@@ -0,0 +1,401 @@
...
+                prefix += part + "/"
+
+        match = re.compile(fnmatch.translate(pattern)).fullmatch
+        res = []
+        for obj_status in await self.list_objects(bucket_name, prefix=prefix):
+            if match(obj_status.name):
+                res.append(obj_status)
+        return res
+
+    async def head_object(self, bucket_name: str, key: str) -> ObjStatus:
```
tvoinarovskyi: The APIs here are explicitly getting `bucket_name`. Thus when calling we end up with:
```python
head_object(bucket_name=uri.host, key=uri.path.lstrip('/'))
```
Another way to do it would be to get an URL object and parse it here. Maybe it would be a better API.

neuromation/api/object_storage.py
```diff
@@ -0,0 +1,401 @@
...
+                    modification_time = int(time.mktime(timetuple))
+            return ObjStatus(
+                path=key,
+                type=type_,
+                size=resp.content_length or 0,
+                modification_time=modification_time,
+                bucket_name=bucket_name,
+            )
+
+    async def fetch_object(self, bucket_name: str, key: str) -> AsyncIterator[bytes]:
```
tvoinarovskyi: GetObject in botocore returns the stream as part of the response. I was hesitant to return an object that would take responsibility for yet not closed request `Object(resp=resp)`. But the response headers and metadata are not accessible using this API, which bothers me.

neuromation/api/object_storage.py
tvoinarovskyi: Or maybe try to do similar signature with Storage with `stat` and `open`

asvetlov: The file storage operates with `URL` instances.
Maybe, the same abstraction is good for object storage as well.

neuromation/api/object_storage.py
```diff
@@ -0,0 +1,524 @@
...
+
+async def calc_md5(path: Path) -> str:
+    loop = asyncio.get_event_loop()
+    md5 = hashlib.md5()
+    with path.open("rb") as stream:
+        while True:
+            chunk = await loop.run_in_executor(None, stream.read, READ_SIZE)
+            if not chunk:
+                break
+            md5.update(chunk)
```
asvetlov: The whole loop should be run in the executor I guess. `md5.update()` is CPU-bound operation.

neuromation/api/object_storage.py
asvetlov: In this case I can use `object_status.uri`, right?

neuromation/api/object_storage.py
asvetlov: Does `Tuple[ObjectStatus, AsyncIterator[bytes]]` solve your needs?
Or maybe a dataclass instead of tuple?

neuromation/api/object_storage.py
tvoinarovskyi: Yes, that is correct.

neuromation/api/abc.py
```diff
@@ -9,42 +9,46 @@
 # storage
 
 
+class StorageProgressEvent:
```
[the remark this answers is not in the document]
tvoinarovskyi: Data type declarations

neuromation/api/config.py
```diff
@@ -198,6 +198,12 @@ def admin_url(self) -> URL:
         # API URL prefix: api/v1, admin prefix: apis/admin/v1
         return self._config_data.url.parent.parent / "apis" / "admin" / "v1"
 
+    @property
+    def obj_url(self) -> URL:
+        cluster = self._config_data.clusters[self._config_data.cluster_name]
+        # XXX: Updata properly the API to return Object storage as separate URL
+        return URL(str(cluster.storage_url).replace("/storage", "/obs"))
```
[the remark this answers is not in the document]
tvoinarovskyi: Yes, it should. But that is 2 PR's away. Will update it as soon as that is finished.

neuromation/cli/main.py
tvoinarovskyi: In this case, I hide the commands until we release it. Don't want to halt a release if needed, as servers are yet to be provisioned with object storage. 
What would be the behaviour of CLI commands if Object Storage is not supported for this cluster? Just an error when you try to use it?

neuromation/api/object_storage.py
```diff
@@ -0,0 +1,524 @@
...
+MAX_OPEN_FILES = 20
+READ_SIZE = 2 ** 20  # 1 MiB
+
+ProgressQueueItem = Optional[StorageProgressEvent]
+
+
+# We extend from FileStatus to make sure our formatting system from Storage can be
+# reused
+@dataclass(frozen=True)
+class ObjStatus(FileStatus):
```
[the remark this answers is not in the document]
tvoinarovskyi: You are correct. Let me see how we can handle formatters. Probably updating formatters to 
Union[FileStatus, ObjectStatus] and using duck-typing should be a more graceful solution

neuromation/api/object_storage.py
tvoinarovskyi: That is correct, yes.

neuromation/api/client.py
```diff
@@ -97,6 +99,10 @@ def jobs(self) -> Jobs:
     def storage(self) -> Storage:
         return self._storage
 
+    @property
+    def obj(self) -> ObjectStorage:
```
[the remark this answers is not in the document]
tvoinarovskyi: `object_storage`?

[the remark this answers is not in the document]
tvoinarovskyi: It's not too long anyway.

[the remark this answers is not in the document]
tvoinarovskyi: or we could make the terminology Blob Storage. Thus `blob://`, `neuro blob ls`, `client.blobs.ls()`

tvoinarovskyi:
Close in favor of #1438

PR #1393, opened 2020-03-12
Job tags support

ayushkovskiy:


neuromation/cli/job.py
```diff
@@ -557,13 +575,17 @@ def job() -> None:
     neuro ps --name my-experiments-v1 -s failed -s succeeded
     neuro ps --description="my favourite job"
     neuro ps -s failed -s succeeded -q
+    neuro ps --tag tag1 -t tag2
```
asvetlov: Let's use the only short or long version in the example (doesn't matter what form will you choose).
I believe it is less confusing, people can learn an alternative by `neuro help ps` easily.

ayushkovskiy: will consider in the following-up PR.

PR #1395, opened 2020-03-13
Use multiple users for e2e tests

asvetlov:


tests/e2e/test_e2e_admin.py
```diff
@@ -1,14 +1,28 @@
...
 import pytest
 
 from tests.e2e import Helper
 
 
+pytestmark = pytest.mark.require_admin
+
+
+@pytest.mark.e2e
+def test_list_clusters(helper: Helper) -> None:
```
serhiy-storchaka: These two tests do not require admin. Maybe left them in a different file?

asvetlov: Ok, let me mark every test explicitly then.

PR #1401, opened 2020-03-16
File expand on relative paths with pattern

tvoinarovskyi:
The problem:
```
⇒  neuro cp "~/Downloads/DEMO/*" storage:
Copy file:///Users/taras/Downloads/DEMO/* => storage://tvoinarovskyi/*
ERROR: cannot copy file:///Users/taras/Downloads/DEMO/* to storage://tvoinarovskyi/*: [Errno 2] No such file: '/Users/taras/Downloads/DEMO/*'
```

Added tests for _expand on file schemas too

### DIFF
--- a/neuromation/cli/storage.py
+++ b/neuromation/cli/storage.py
@@ -5,6 +5,7 @@ import os
 import secrets
 import shlex
 import sys
+from pathlib import Path
 from typing import Any, Callable, List, Optional, Sequence, Tuple
 
 import aiodocker
@@ -844,7 +845,7 @@ async def _expand(
                     uris.append(file)
             elif allow_file and uri.scheme == "file":
                 for p in globmodule.iglob(uri_path, recursive=True):
-                    uris.append(uri.with_path(p))
+                    uris.append(URL(Path(p).as_uri()))
             else:
                 uris.append(uri)
         else:
--- a/neuromation/cli/storage.py
+++ b/neuromation/cli/storage.py
@@ -842,7 +842,7 @@ async def _expand(
             if uri.scheme == "storage":
                 async for file in root.client.storage.glob(uri):
                     uris.append(file)
-            elif allow_file and path.startswith("file:"):
+            elif allow_file and uri.scheme == "file":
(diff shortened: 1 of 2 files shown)


serhiy-storchaka:
I did not enable this at first place because it may conflict with path expansions in the shell. If you have a name containing `*` in your `~/Downloads/DEMO/` directory
```
neuro cp Downloads/DEMO/* storage:
```
the argument will first be expanded by the shell, and then by `neuro`.

tvoinarovskyi:
@serhiy-storchaka I did think that was the case, but it works fine with shell expand even with the fix. It would be great to have it working in cases where we launch the command with no shell or with quoted parameters.

serhiy-storchaka:
```
mkdir data
touch 'data/*.txt'
touch data/abc.txt
neuro -v cp data/?.txt storage:
```
Currently it only copies a single file. With your change it will copy all files.

tvoinarovskyi:
@serhiy-storchaka Ok, I see your point. And I presume other software would also have this problem... For example:
```
taras@Tarass-MacBook-Pro:~/workspace/platform-client-python|fix_relative_pattern⚡
⇒  gsutil cp data/?.txt gs://neuro-test-gcs/
Copying file://data/abc.txt [Content-Type=text/plain]...
Copying file://data/*.txt [Content-Type=text/plain]...
/ [2 files][    0.0 B/    0.0 B]
Operation completed over 2 objects.
taras@Tarass-MacBook-Pro:~/workspace/platform-client-python|fix_relative_pattern
⇒  ls -l data/?.txt
-rw-r--r--  1 taras  staff  0 Mar 17 14:25 data/*.txt
taras@Tarass-MacBook-Pro:~/workspace/platform-client-python|fix_relative_pattern
⇒
```
Lets document the behaviour then. I feel like it's OK to state, that glob patterns will only be resolved if used in URI form.

serhiy-storchaka:
We could also enable it on Windows where patterns are not expanded by a shell. Or use other heuristic. But I think it is safer to use a simple rule -- to expand glob patterns only in URIs. We can change it in future.

PR #1402, opened 2020-03-16
Simplify progress queue in Storage a little

tvoinarovskyi:
I found the queue containing only a single object and None to specify closure to work much simpler. Thoughts?

### DIFF
--- a/neuromation/api/storage.py
+++ b/neuromation/api/storage.py
@@ -20,6 +20,7 @@ from typing import (
     List,
     Optional,
     Tuple,
+    cast,
 )
 
 import aiohttp
@@ -31,6 +32,7 @@ from .abc import (
     AbstractRecursiveFileProgress,
     StorageProgressComplete,
     StorageProgressEnterDir,
+    StorageProgressEvent,
     StorageProgressFail,
     StorageProgressLeaveDir,
     StorageProgressStart,
@@ -52,7 +54,7 @@ READ_SIZE = 2 ** 20  # 1 MiB
 TIME_THRESHOLD = 1.0
 
 Printer = Callable[[str], None]
-ProgressQueueItem = Optional[Tuple[Callable[[Any], None], Any]]
+ProgressQueueItem = Optional[StorageProgressEvent]
 
 
 class FileStatusType(str, enum.Enum):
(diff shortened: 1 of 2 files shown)


neuromation/api/storage.py
```diff
@@ -749,29 +692,27 @@ def _file_status_from_api(values: Dict[str, Any]) -> FileStatus:
...
 
-    def step(self, data: StorageProgressStep) -> None:
-        pass
-
-    def enter(self, data: StorageProgressEnterDir) -> None:
-        pass
+        if progress is None:
+            # Just ignore event if we don't have a listener
+            continue
+        if isinstance(item, StorageProgressStart):
```
serhiy-storchaka: Maybe replace all this with
```python
getattr(progress, item.methodname)(item)
```
?

asvetlov: No, please don't store a name in an item's attribute.

tvoinarovskyi: It kind of feels like a nice single dispatch use case. For example progress.push(), where push would dispatch

serhiy-storchaka: I propose to save it as a class attribute.

Alternatively we can add a lot of trivial methods like
```python
class StorageProgressStart:
    def doit(self, progress: AbstractFileProgress) -> None:
        progress.start(self)
```
But it is a lot of repetitions. I think that specifying a method name is more Pythonic.
```python
class StorageProgressStart:
    methodname = "start"
```

asvetlov: There is another alternative: drop `AbstractFileProgress` eventually for the sake of async iterator based approach:
```
async for msg in client.storage.upload_dir(src, dst):
    if type(msg) is StorageProgressStart:
        ....
    elif type(msg) is StorageProgressComplete:
        ....
```
It allows to implement all progress-related operations in the signle functions, using local variables instead of instance vars for troring the state.

This approach is used by https://github.com/python-hyper/hyper-h2 and family.

tvoinarovskyi: @asvetlov That approach scales not very good with file upload, as it needs to spawn many coroutines to process files. It's easier to push into a queue from those worker tasks and process it with a single coroutine. Of course, we could change that last task to a generator and yield those messages into cli, so that we don't need AbstractFileProgress, but that does not solve the dispatch problem Serhiy is proposing to change. You will still end up with dispatching using several if statements. 
Personally I only wanted to avoid passing 2 variables "progress" and "queue" around. I don't see any big problem with either dispatching using many if statements nor adding a "type" or "method" to events to make it more speedy, as it does not change the upload code and leaves it relatively easy to work with.

serhiy-storchaka: @asvetlov I like this idea, but it is not compatible with `_iterate_file()` which is aready an iterator yielding chunks of data.

Alternatively we could introduce a progress class for which `await progress.start(...)` would just put an event and the corresponding method of the underlying progress object (`progress._internal_progress.start`) into an internal queue.

tvoinarovskyi: @serhiy-storchaka Tried the last approach here https://github.com/neuromation/platform-client-python/compare/simplify_progress_v2?expand=1. It works, but in the end I feels like it's confusing to have 2 Progress interfaces... This PR sits better with me.

serhiy-storchaka: Could you create a PR for it?

tvoinarovskyi:
Fixed by #1420

Issue #1405, opened 2020-03-17
Display job's tags in neuro status


ayushkovskiy:
Tech debt

serhiy-storchaka:
It the column for tags available in `neuro ps`?

ayushkovskiy:
Thanks @serhiy-storchaka ! I'll do my best to do this until the next client release.

PR #1407, opened 2020-03-17
Display job's tags in neuro status

ayushkovskiy:
Closes https://github.com/neuromation/platform-client-python/issues/1405

### DIFF
--- a/neuromation/cli/formatters/jobs.py
+++ b/neuromation/cli/formatters/jobs.py
@@ -78,6 +78,8 @@ class JobStatusFormatter:
         result: str = f"Job: {job_status.id}\n"
         if job_status.name:
             result += f"Name: {job_status.name}\n"
+        if job_status.tags:
+            result += f"Tags: {', '.join(job_status.tags)}\n"
         result += f"Owner: {job_status.owner if job_status.owner else ''}\n"
         result += f"Cluster: {job_status.cluster_name}\n"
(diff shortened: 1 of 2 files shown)


neuromation/cli/formatters/jobs.py
```diff
@@ -78,6 +78,8 @@ def __call__(self, job_status: JobDescription) -> str:
         result: str = f"Job: {job_status.id}\n"
         if job_status.name:
             result += f"Name: {job_status.name}\n"
+        if job_status.tags:
+            result += f"Tags: {', '.join(job_status.tags)}\n"
```
ayushkovskiy: Maximum number of the job's tags is `256`, max length of each tag is `256`. I don't think it's a problem to have a very-long line in the worst case, but if anyone has concerns, we could think how to separate this value to multiple lines.

serhiy-storchaka: It is easy.
```python
text = "Tags: " + ", ".join(job_status.tags)
if root.tty:
    width = root.get_terminal_size[0]
    text = textwrap.fill(text, width=width, subsequent_indent="      ", break_long_words=False)
result += text +"\n"
```

ayushkovskiy: Thank you. This is the issue not to forget: https://github.com/neuromation/platform-client-python/issues/1408

Issue #1406, opened 2020-03-17
Display job's tags in neuro ps


ayushkovskiy:
Tech debt, https://github.com/neuromation/platform-client-python/issues/1405#issuecomment-600146481

PR #1410, opened 2020-03-18
Display tags in `neuro ps`

ayushkovskiy:
Optionally: only if `.neuro.toml` config has explicit `tags` section.

Closes https://github.com/neuromation/platform-client-python/issues/1406

### DIFF
--- a/neuromation/cli/topics.py
+++ b/neuromation/cli/topics.py
@@ -111,10 +111,10 @@ async def ps_format() -> None:
       for indication of alignment, minimum and maximum column width,
       and optional column title:
 
-        `{id;align=center;min=10;max=30;width=20;ID TITLE}`
+        `{id;align=center;min=10;max=30;width=20;TITLE}`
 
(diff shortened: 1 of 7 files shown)


neuromation/cli/topics.py
```diff
@@ -111,10 +111,10 @@ def topics() -> None:
       for indication of alignment, minimum and maximum column width,
       and optional column title:
 
-        `{id;align=center;min=10;max=30;width=20;ID TITLE}`
+        `{id;align=center;min=10;max=30;width=20;TITLE}`
```
ayushkovskiy: Side change: `ID TITLE` (with space) is confusing.

ayushkovskiy:
I didn't find any test on `neuro ps` output, so no new test for current functionality provided. Does this PR need a changelog note?

serhiy-storchaka:
Yes, I think it needs a changelog note. Otherwise how users can know about it?

ayushkovskiy:
Added changelog.

PR #1411, opened 2020-03-18
Extract click types to a separate module

asvetlov:


serhiy-storchaka:
`types` looks too general. Maybe `click_types`?

asvetlov:
Agree, `click_types` looks better.

PR #1417, opened 2020-03-20
One page per command group

mariyadavydova:


mariyadavydova:
@asvetlov I would also like to generate topics, but I couldn't find a way to extract them from Click. Can you please assist?

asvetlov:
LGTM.
Sorry, the topics generator is not easy.
The text contains some markup but it uses a custom syntax that is not compatible with markdown.
The output looks ugly, you can check it by uncommenting `gen_topics()` call in the `main()` function.
Theoretically, the custom  syntax can be adopted to markdown but this task requires up to a few days for implementing the support of tables, code sections etc.

PR #1431, opened 2020-03-25
Add option to pass pytest additional arguments to make

tvoinarovskyi:
@asvetlov I just very like to run tests as such:
```
make test PYTEST_ARGS="-k test_object_storage -vv"
```


asvetlov:
Up to you.
I personally just use `pytest -k test_object_storage -vv` for your case.

serhiy-storchaka:
I also just use pytest.

PR #1435, opened 2020-03-26
storage tree command

asvetlov:


neuromation/cli/storage.py
```diff
@@ -835,6 +838,67 @@ def filter_option(*args: str, flag_value: bool, help: str) -> Callable[[Any], An
...
+    """
+    if not path:
+        path = "storage:"
+    uri = parse_file_resource(path, root)
+
+    errors = False
+    try:
+        tree = await fetch_tree(root.client, uri, show_all)
+        tree = dataclasses.replace(tree, name=str(path))
+    except (OSError, ResourceNotFound, IllegalArgumentError) as error:
```
serhiy-storchaka: I am not sure that `IllegalArgumentError` is still needed. It was raised when you tried to get a listdir of a non-directory. Now it should raise `OSError` or `ResourceNotFound`.

asvetlov: Ok, removed.

asvetlov:
Currently, the pager (from `click`) doesn't work well with async generators.
Also, it requires the whole content before starting paging on Windows.
There is an ability to reimplement the pager from scratch to support all features.
Maybe I'll find a time for it but definitely now I have tasks with higher priority.

PR #1440, opened 2020-03-27
fix external alias optional args parsing

zubenkoivan:


tests/cli/test_alias.py
```diff
@@ -475,7 +475,7 @@ def test_external_alias_three_args_optional_multiple_regular_help(
         prog_name = Path(sys.argv[0]).name
         expected = inspect.cleandoc(
             f"""\
-            Usage: {prog_name} user-cmd [OPTIONS] [ARG1] ARG2... ARG3
+            Usage: {prog_name} user-cmd [OPTIONS] ARG1 ARG2... [ARG3]
```
zubenkoivan: Args parsing was working correctly only when optional arg was in first position, otherwise `Missing close bracket` was raised

asvetlov:
thanks!

PR #1441, opened 2020-03-30
Changes to Blob Storage API

tvoinarovskyi:
Add bucket create/delete APIs
Change URL path to use cluster_name
Add permissions to list bucket results

### DIFF
--- a/neuromation/api/blob_storage.py
+++ b/neuromation/api/blob_storage.py
@@ -342,7 +342,8 @@ class BlobStorage(metaclass=NoPublicConstructor):
 
     async def _mkdir(self, uri: URL) -> None:
         bucket_name, key = self._extract_bucket_and_key(uri)
-        assert key.endswith("/")
+        assert key.endswith("/"), "Key should end with a trailing slash"
+        assert key.strip("/"), "Can not create a bucket root folder"
         await self.put_blob(bucket_name=bucket_name, key=key, body=b"")
 
     def make_url(self, bucket_name: str, key: str) -> URL:
@@ -451,17 +452,20 @@ class BlobStorage(metaclass=NoPublicConstructor):
         if not dst.path.endswith("/"):
             dst = dst / ""
 
-        # Make sure we don't have name conflicts
         bucket_name, key = self._extract_bucket_and_key(dst)
-        try:
-            # We can't upload to folder `/path/to/file.txt/` if `/path/to/file.txt`
-            # already exists
-            await self.head_blob(bucket_name=bucket_name, key=key.rstrip("/"))
-        except ResourceNotFound:
-            pass
-        else:
-            raise NotADirectoryError(errno.ENOTDIR, "Not a directory", str(dst))
-        await self._mkdir(dst)
+        if key.rstrip("/"):
+            try:
+                # Make sure we don't have name conflicts
+                # We can't upload to folder `/path/to/file.txt/` if `/path/to/file.txt`
+                # already exists
+                await self.head_blob(bucket_name=bucket_name, key=key.rstrip("/"))
+            except ResourceNotFound:
+                pass
+            else:
+                raise NotADirectoryError(errno.ENOTDIR, "Not a directory", str(dst))
+
+            # Only create folder if we are not uploading to bucket root
+            await self._mkdir(dst)
 
         await progress.enter(StorageProgressEnterDir(src, dst))
         loop = asyncio.get_event_loop()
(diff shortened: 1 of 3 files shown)


neuromation/api/blob_storage.py
```diff
@@ -301,6 +314,18 @@ def __init__(self, core: _Core, config: Config) -> None:
...
                 await progress.complete(StorageProgressComplete(src_url, dst, size))
 
+    def _extract_bucket_and_key(self, uri: URL) -> Tuple[str, str]:
+        cluster_name = self._config.cluster_name
+        uri = normalize_blob_path_uri(uri, cluster_name)
+        if uri.host != self._config.cluster_name:
+            raise ValueError(
+                f"When using full URL's please specify cluster name "
+                f"{cluster_name!r} as host part. For example: "
+                f"blob://{cluster_name!r}/my_bucket/path/to/file."
```
serhiy-storchaka:
```suggestion
                f"blob://{cluster_name}/my_bucket/path/to/file."
```

neuromation/api/blob_storage.py
```diff
@@ -309,30 +334,28 @@ def __init__(self, core: _Core, config: Config) -> None:
             return True
         # Check if a folder key exists. As `/` at the end makes a different key, make
         # sure we ask for one with ending slash.
-        key = _extract_key(uri) + "/"
```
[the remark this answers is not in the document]
tvoinarovskyi: no

neuromation/api/blob_storage.py
```diff
@@ -309,30 +334,28 @@ def __init__(self, core: _Core, config: Config) -> None:
...
+            bucket_name=bucket_name, prefix=key + "/", recursive=False, max_keys=1
         )
         return bool(blobs)
 
     async def _mkdir(self, uri: URL) -> None:
-        assert uri.host
-        assert uri.path.endswith("/")
-        await self.put_blob(bucket_name=uri.host, key=_extract_key(uri), body=b"")
+        bucket_name, key = self._extract_bucket_and_key(uri)
+        assert key.endswith("/")
```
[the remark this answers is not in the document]
tvoinarovskyi: Should not be and will trigger this exception. Let me check if it's possible to trigger this.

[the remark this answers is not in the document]
tvoinarovskyi: Nice find! It actually does trigger this. Let me do a test really quick.

neuromation/api/blob_storage.py
tvoinarovskyi: Thanks!

neuromation/api/blob_storage.py
```diff
@@ -51,11 +51,10 @@
 
 
 def _format_bucket_uri(bucket_name: str, key: str = "") -> URL:
-    return URL.build(scheme="blob", host=bucket_name, path="/" + key.lstrip("/"))
-
-
-def _extract_key(uri: URL) -> str:
-    return uri.path.lstrip("/")
+    if key:
+        return URL(f"blob:{bucket_name}/{key}")
```
serhiy-storchaka: Not sure is this is relevant, but just for the case if `bucket_name` or `key` contain something like `%20` I would use `URL.build()`.

tvoinarovskyi: I was not successful with URL.build producing a relative URL =(. I do think it's a good idea.
```python
>>> URL.build(scheme="blob", path="folder/key.txt")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/taras/.pyenv/versions/3.7.4/lib/python3.7/site-packages/yarl/__init__.py", line 202, in build
    raise ValueError('Can\'t build URL with "scheme" but without "host".')
ValueError: Can't build URL with "scheme" but without "host".
```
I will try to operate on full URLs. Give me a sec.

tvoinarovskyi: Hmm, this seem to work:
```python
>>> yarl.URL("blob:") / "foo" / "мрія"
URL('blob:foo/%D0%BC%D1%80%D1%96%D1%8F')
```

tvoinarovskyi: Hmm, strangely the %20 case does not actually work:
```python
>>> yarl.URL.build(scheme="blob", host="default", path="/foo/%20object1")
URL('blob://default/foo/%20object1')
>>> yarl.URL.build(scheme="blob", host="default", path="/foo/ object1")
URL('blob://default/foo/%20object1')
```
It does not requote the first one.

serhiy-storchaka: It may be a bug in `yarl`.

I like `URL("blob:") / ...`.

PR #1443, opened 2020-03-31
Make sure glob does not consume recursive directories

tvoinarovskyi:
I know the fix is a bit hacky, but can't find a better way for now. Tried to do a recursive iteration, as it's done in storage but it did not perform very well, it needs some attention to do it right. What do you think of this fast fix for now?

This is the related function in gsutil:
https://github.com/GoogleCloudPlatform/gsutil/blob/master/gslib/wildcard_iterator.py
Quite a lot of edge cases as it seems.

### DIFF
--- a/neuromation/api/blob_storage.py
+++ b/neuromation/api/blob_storage.py
@@ -278,7 +278,7 @@ class BlobStorage(metaclass=NoPublicConstructor):
             return
 
         has_magic = _has_magic(part)
-        # Optimize the prefix for matchin. If we have a pattern `folder1/b*/*.json`
+        # Optimize the prefix for matching. If we have a pattern `folder1/b*/*.json`
         # it's better to scan with prefix `folder1/b` on the 2nd step, not `folder1/`
         if has_magic:
             opt_prefix = prefix + _glob_safe_prefix(part)
--- a/neuromation/api/blob_storage.py
+++ b/neuromation/api/blob_storage.py
@@ -17,6 +17,7 @@ from typing import (
     Dict,
     List,
     Optional,
+    Sequence,
     Tuple,
     Union,
     cast,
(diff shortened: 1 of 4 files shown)


asvetlov:
Do you have a thought about how it can be implemented in an iterative manner without fetching everything in memory?
A number of objects in bucket is unlimited as far as I know.

serhiy-storchaka:
What exactly problem it solves?

tvoinarovskyi:
@serhiy-storchaka So the problem case is as follows:
```
neuro blob glob *.jpg
```
will return all subdirectories, while I would expect it to return only 1 folder deep.

tvoinarovskyi:
@asvetlov Yes I do have thoughts, but will require to spend a good 2 days implementing, so I proposed this hack to handle the case for now.

tvoinarovskyi:
Relates to https://github.com/neuromation/platform-client-python/pull/1445

neuromation/api/blob_storage.py
```diff
@@ -217,6 +218,7 @@ def __init__(self, core: _Core, config: Config) -> None:
     async def glob_blobs(self, bucket_name: str, pattern: str) -> List[BlobListing]:
         pattern = pattern.lstrip("/")
         parts = pattern.split("/")
```
serhiy-storchaka: Humm, the pattern can contain something like `[-/]`.

neuromation/api/blob_storage.py
```diff
@@ -225,7 +227,40 @@ def __init__(self, core: _Core, config: Config) -> None:
...
+                for matcher, part in itertools.zip_longest(matchers, parts):
+                    if part is None or matcher is None:
+                        return False
+                    if not matcher(part):
+                        return False
+                return True
+
+        else:
+            matcher = "/".join(matcher_parts) + r"\Z"
+            match = re.compile(matcher).fullmatch
```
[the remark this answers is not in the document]
tvoinarovskyi: Oh, yea...

neuromation/api/blob_storage.py
tvoinarovskyi: Then it would be an invalid URL.

serhiy-storchaka: It is not a URL, it is a pattern. `foo[-/]bar` matches two paths: `foo-bar` and `foo/bar`.

tvoinarovskyi: Is that a valid glob case?

PR #1445, opened 2020-03-31
Make filter patterns more compatible with .gitignore.

### LINKED ISSUE #1444: Implement more standard filter patterns
There are several issues with filters:

* `?` matches `/`. So `a?c` matches `abc` as well as `a/c`.
* `*` matches sequences containing `/`. So `a*z` matches `abz`, `a/z`, `ab/yz`, `ab/c/yz` etc.
* To match files in subdirectories you have to prepend `*/`: `--exclude '*.tmp' --exclude '*/*.tmp'`.

We have to implements semantic used in `.gitignore`.

* Make `?` and `*` not matching `/`.
* Add support of `**` for recursive matching.
* Make patterns not containing slashes except the trailing slashes recursive (virtually prepend implicit `**/` to them).

serhiy-storchaka:
* Make `?` and `*` not matching `/`.
* Add support of `**` for recursive matching.
* Make patterns not containing slashes except the trailing slashes recursive (virtually prepend implicit `**/` to them).

Closes #1444.

### DIFF
--- a/neuromation/api/file_filter.py
+++ b/neuromation/api/file_filter.py
@@ -1,4 +1,3 @@
-import fnmatch
 import re
 from typing import Any, Callable, List, Tuple, cast
 
@@ -8,8 +7,12 @@ class FileFilter:
         self.filters: List[Tuple[bool, Callable[[str], Any]]] = []
 
     def append(self, exclude: bool, pattern: str) -> None:
-        re_pattern = fnmatch.translate(pattern)
-        matcher = cast(Callable[[str], Any], re.compile(re_pattern).match)
+        if "/" not in pattern.rstrip("/"):
+            pattern = "**/" + pattern
+        re_pattern = translate(pattern)
+        matcher = cast(
+            Callable[[str], Any], re.compile(re_pattern, re.DOTALL).fullmatch
+        )
         self.filters.append((exclude, matcher))
 
     def exclude(self, pattern: str) -> None:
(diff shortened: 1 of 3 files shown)


neuromation/api/file_filter.py
```diff
@@ -24,3 +27,76 @@ def include(self, pattern: str) -> None:
...
+                    res += ".*"
+                else:
+                    res += "(?:.+/)?"
+                i += 2
+            else:
+                # Any other *
+                res += "[^/]*"
+        elif c == "?":
+            res += "[^/]"
+        elif c == "/":
```
serhiy-storchaka, line 60: Special case for `/` was added just to make tests easier: in 3.6 `/` is escaped by `re.escape()`, in 3.7 it is not escaped.

neuromation/api/file_filter.py
```diff
@@ -24,3 +27,76 @@ def include(self, pattern: str) -> None:
...
+            res += "/"
+        elif c == "[":
```
serhiy-storchaka, line 62: The rest of the code is virtually identical to `fnmatch.translate()`.

tests/api/test_file_filter.py
```diff
@@ -33,3 +33,71 @@
     assert await ff.match("spam.txt")
     assert await ff.match("ham")
     assert not await ff.match("ham.txt")
+    assert await ff.match("dir/spam.txt")
+    assert not await ff.match("dir/ham.txt")
+    assert await ff.match("dir.txt/spam")
```
serhiy-storchaka, line 38: There is a difference with `.gitignore` for this case. Humm, I thought it is not significant, but now I think that it may be important.

serhiy-storchaka: Well, it is not important in the context of using filters, because we handle files recursively and omit the content of a directory if its name is filtered out.

tvoinarovskyi: I think this is fine, as it is specific to filter, not translate.

PR #1451, opened 2020-04-01
Minor changes in handling resource URIs.

serhiy-storchaka:
* Require absolute URI for schemes other than "storage", "image", "job" and "blob".
* Allow sharing resources "blob" and "role".


tvoinarovskyi:
@serhiy-storchaka Can a user create roles somehow? That sounds like a cool feature. Right now he can possibly share `his` role with other people in theory, right?

PR #1457, opened 2020-04-06
Use streamed format for storage listing.

serhiy-storchaka:
`Storage.ls()` is an asynchronous generator now. Use
```python
async for status in storage.ls(...):
```
instead of
```python
for status in await storage.ls(...):
```


asvetlov:
May I ask you to update API docs to reflect the change?

serhiy-storchaka:
#1463

PR #1475, opened 2020-04-16
Remove bucket dependency from Helper fixture.

tvoinarovskyi:
Only blob storage tests should create temporary bucket

### DIFF
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -96,14 +96,11 @@ def run_async(coro: Any) -> Callable[..., Any]:
 
 
 class Helper:
-    def __init__(
-        self, nmrc_path: Optional[Path], tmp_path: Path, tmpbucketname: str
-    ) -> None:
+    def __init__(self, nmrc_path: Optional[Path], tmp_path: Path) -> None:
         self._nmrc_path = nmrc_path
         self._tmp = tmp_path
         self.tmpstoragename = f"test_e2e/{uuid()}"
         self._tmpstorage = f"storage:{self.tmpstoragename}/"
-        self._tmpbucketname = tmpbucketname
         self._closed = False
         self._executed_jobs: List[str] = []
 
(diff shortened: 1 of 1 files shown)


tests/e2e/conftest.py
```diff
@@ -689,15 +681,25 @@ def nested_data(static_path: Path) -> Tuple[str, str, str]:
 
 
 @pytest.fixture(scope="session")
-def tmp_bucket(tmp_path_factory: Any, request: Any) -> Iterator[str]:
+def tmp_bucket_create(
```
asvetlov, line 684: Hint: `pytest` assumes that fixtures started with an underscore are *private*, they are not displayed by `pytest --fixtures` etc.
Maybe this is a good convention for `tmp_bucket_create`?

asvetlov, line 684:
```suggestion
def _tmp_bucket_create(
```

tests/e2e/conftest.py
```diff
@@ -689,15 +681,25 @@ def nested_data(static_path: Path) -> Tuple[str, str, str]:
...
-    helper.delete_bucket(helper.tmpbucketname)
+    helper = Helper(nmrc_path, tmp_path)
+    helper.create_bucket(tmpbucketname)
+    yield tmpbucketname, helper
+    helper.delete_bucket(tmpbucketname)
+    helper.close()
+
+
+@pytest.fixture
+def tmp_bucket(tmp_bucket_create: Tuple[str, Helper]) -> Iterator[str]:
```
asvetlov, line 699:
```suggestion
def tmp_bucket(_tmp_bucket_create: Tuple[str, Helper]) -> Iterator[str]:
```

tests/e2e/conftest.py
```diff
@@ -689,15 +681,25 @@ def nested_data(static_path: Path) -> Tuple[str, str, str]:
...
+    tmpbucketname, helper = tmp_bucket_create
```
asvetlov, line 700:
```suggestion
    tmpbucketname, helper = _tmp_bucket_create
```

PR #1476, opened 2020-04-17
Check blob permission and skip test if none

tvoinarovskyi:
If the configured user running tests does not have permission to create a bucket (and by default users don't) we can just skip tests for him instead of raising a bunch of errors.

### DIFF
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -31,6 +31,7 @@ import pytest
 from yarl import URL
 
 from neuromation.api import (
+    AuthorizationError,
     Config,
     Container,
     FileStatusType,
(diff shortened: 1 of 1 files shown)


tests/e2e/conftest.py
```diff
@@ -686,10 +687,15 @@ def _tmp_bucket_create(
...
-    nmrc_path = _get_nmrc_path(tmp_path_factory, require_admin=False)
+    nmrc_path = _get_nmrc_path(tmp_path_factory, require_admin=True)
 
     helper = Helper(nmrc_path, tmp_path)
-    helper.create_bucket(tmpbucketname)
+    try:
+        helper.create_bucket(tmpbucketname)
+    except AuthorizationError:
+        pytest.skip("No permission to create bucket for user E2E_TOKEN")
+        yield "", helper
```
asvetlov, line 697: `pytest.skip()` raises an exception internally, this line is never executed.

tvoinarovskyi: Ohhh, is that so. Let me check if I can remove it then.

PR #1485, opened 2020-04-22
autocomplete job name/id

asvetlov:


neuromation/cli/click_types.py
```diff
@@ -193,3 +194,41 @@ async def async_complete(
...
+                job_name = job.name or ""
+                for test in (
+                    job.id,
+                    job_name,
+                    f"job:{job.id}",
+                    f"job:/{client.username}/{job.id}",
+                    f"job://{client.cluster_name}/{client.username}/{job.id}",
+                ):
+                    if test.startswith(incomplete):
+                        ret.append((test, job_name))
```
serhiy-storchaka, line 229: What is the meaning of this tuple? I have not find the documentation.

asvetlov: The first item is the found match, the second is for help string (used in `zsh` completion only).

neuromation/cli/click_types.py
```diff
@@ -193,3 +194,41 @@ async def async_complete(
...
+            now = datetime.now()
+            async for job in client.jobs.list(
+                since=now - timedelta(days=7), reverse=True, limit=100
+            ):
+                job_name = job.name or ""
+                for test in (
+                    job.id,
+                    job_name,
+                    f"job:{job.id}",
+                    f"job:/{client.username}/{job.id}",
```
serhiy-storchaka, line 225: If you use the form `job:/{user-name}/{job-id}`, `user-name` is usually the other user's name. You can use this name for filtering.

If you use the form `job:{job-id}` (without `/` after `job:`), you refer to your own jobs, so you can pass `owners=[client.username]` to `list()`.

neuromation/cli/click_types.py
asvetlov: I've found that just job ID doesn't help too much to select the job.
The help text can contain more info, suggestions are welcome!

serhiy-storchaka: Then you should return `job_name` as the first item if `job_name.startswith(incomplete)` and an URI if `incomplete` matches `job:...`.

neuromation/cli/click_types.py
asvetlov: thanks, now I take cluster and user from job-description.

Regarding `owners=[client.username]` -- I prefer to make the single REST call for the whole autocomplete, it makes code much simpler. Do you have an objection?

serhiy-storchaka: Ok, it works now. Maybe adding additional parameters taken from `incomplete` could make the request faster for admins, but I am not sure the difference will be significant.

But I think that you should return a name or an URI when autocomplete a name or an URI.

serhiy-storchaka: Please ignore the last phrase. I missed that the current code does exactly this.

asvetlov:
Thanks!

PR #1512, opened 2020-05-20
Report Python version and OS name

asvetlov:


neuromation/cli/stats.py
```diff
@@ -130,10 +131,15 @@ async def send(client: Client, uid: str, data: List[sqlite3.Row]) -> None:
         )
         + "\n"
     )
+    neuro_ver = neuromation.__version__
+    plat = sys.platform
+    py_version = ".".join(str(i) for i in sys.version_info[:3])
```
serhiy-storchaka, line 136: Maybe use `platform.python_version()`?

And maybe use `platform.python_implementation()` instead of `"Python"`?

asvetlov: Thanks. Updated

PR #1513, opened 2020-05-20
Refactor RemoteImage API

### LINKED ISSUE #1468: `RemoteImage.registry` is populated only when a Neuro registry is used.
It gets really confusing when one instantiates `RemoteImage` manually and tries to populate constructor parameters. How would one know that if it is a Neuro registry, then `registry` should be populated, but when it is not, then the registry should go into `name`?

If everything goes to `name` then why keep `tag` separately?

anayden:
Closes #1468

### DIFF
--- a/docs/images_reference.rst
+++ b/docs/images_reference.rst
@@ -205,7 +205,7 @@ RemoteImage
 
     .. py:classmethod:: new_external_image(name: str, registry: Optional[str] = None, *, tag: Optional[str] = None) -> RemoteImage
 
-        Create a new instance referring to an image hosted on an external registry (e.g. DockerHub)
+        Create a new instance referring to an image hosted on an external registry (e.g. DockerHub_).
 
       :param str name: name of the image
 
--- a/docs/images_reference.rst
+++ b/docs/images_reference.rst
@@ -191,7 +191,7 @@ RemoteImage
 
     .. py:classmethod:: new_neuro_image(name: str, registry: str, *, owner: str, cluster_name: str, tag: Optional[str] = None) -> RemoteImage
 
-        Create a new instance referring to an image hosted on Neuro Platform
+        Create a new instance referring to an image hosted on Neuro Platform.
 
       :param str name: name of the image
 
--- a/docs/images_reference.rst
+++ b/docs/images_reference.rst
@@ -189,5 +189,29 @@ RemoteImage
 
       URL that can be used to reference this image with Docker.
 
+    .. py:classmethod:: new_neuro_image(name: str, registry: str, *, owner: str, cluster_name: str, tag: Optional[str] = None) -> RemoteImage
+
+        Create a new instance referring to an image hosted on Neuro Platform
+
+      :param str name: name of the image
+
+      :param str registry: registry where the image is located
+
+      :param str owner: image owner name
+
+      :param str cluster_name: name of the cluster
+
+      :param str tag: image tag
+
+    .. py:classmethod:: new_external_image(name: str, registry: Optional[str] = None, *, tag: Optional[str] = None) -> RemoteImage
+
+        Create a new instance referring to an image hosted on an external registry (e.g. DockerHub)
+
+      :param str name: name of the image
+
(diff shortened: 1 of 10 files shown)


neuromation/api/parsing_utils.py
```diff
@@ -33,11 +56,13 @@ def __post_init__(self) -> None:
...
         if _is_in_neuro_registry(self):
-            name = f"https://{self.registry}/{self.owner}/{self.name}"
+            name = f"{self.registry}/{self.owner}/{self.name}"
             tag = f":{self.tag}" if self.tag else ""
             return name + tag
         else:
-            return str(self)
+            name = self.name
+            tag = f":{self.tag}" if self.tag else ""
+            return name + tag
```
serhiy-storchaka, line 65: It is common in both branches.

anayden: good point. I made the changes to simplify the code in the future and forgot about it. Updated

neuromation/api/parsing_utils.py
```diff
@@ -19,6 +19,29 @@ class RemoteImage:
     registry: Optional[str] = None
     cluster_name: Optional[str] = None
 
+    @classmethod
+    def new_neuro_image(
```
serhiy-storchaka, line 23: In `new_external_image()` parameters are: `name`, `tag`, `registry` (the last two are optional).
In `new_neuro_image()` parameters are: `name`, `owner`, `registry`, `cluster_name`, `tag` (only `tag` is optional).
The order of arguments `tag` and `registry` is different, and there are other parameters inserted in the middle.

I propose to move `registry` after `name` and make all parameters except `name` and `registry` keyword-only. In both methods.

neuromation/api/parsing_utils.py
```diff
@@ -32,12 +55,14 @@ def __post_init__(self) -> None:
...
-        else:
-            return str(self)
+        name = self.name
+        name = (
+            f"{self.registry}/{self.owner}/{name}"
+            if _is_in_neuro_registry(self)
+            else name
+        )
+        tag = f":{self.tag}" if self.tag else ""
+        return name + tag
```
serhiy-storchaka, line 65: Maybe so?
```suggestion
        name = self.name
        if _is_in_neuro_registry(self):
            name = f"{self.registry}/{self.owner}/{name}"
        if self.tag:
            name = f"{name}:{self.tag}"
        return name
```
Or so?
```suggestion
        prefix = f"{self.registry}/{self.owner}/" if _is_in_neuro_registry(self) else ""
        suffix = f":{self.tag}" if self.tag else ""
        return f"{prefix}{self.name}{suffix}"
```

docs/images_reference.rst
```diff
@@ -189,5 +189,29 @@ RemoteImage
 
       URL that can be used to reference this image with Docker.
 
+    .. py:classmethod:: new_neuro_image(name: str, registry: str, *, owner: str, cluster_name: str, tag: Optional[str] = None) -> RemoteImage
+
+        Create a new instance referring to an image hosted on Neuro Platform
```
serhiy-storchaka, line 194:
```suggestion
        Create a new instance referring to an image hosted on Neuro Platform.
```

docs/images_reference.rst
```diff
@@ -189,5 +189,29 @@ RemoteImage
...
+
+      :param str owner: image owner name
+
+      :param str cluster_name: name of the cluster
+
+      :param str tag: image tag
+
+    .. py:classmethod:: new_external_image(name: str, registry: Optional[str] = None, *, tag: Optional[str] = None) -> RemoteImage
+
+        Create a new instance referring to an image hosted on an external registry (e.g. DockerHub)
```
serhiy-storchaka, line 208:
```suggestion
        Create a new instance referring to an image hosted on an external registry (e.g. DockerHub_).
```

anayden: I assume, you only suggested to add a dot at the end. Will commit this manually.

anayden: Oh, nevermind, I figured it out.

PR #1516, opened 2020-05-25
Fix setting the tty flag from server response.

serhiy-storchaka:


serhiy-storchaka:
There are no tests yet.

asvetlov:
Two lines here: https://github.com/neuromation/platform-client-python/blob/master/tests/e2e/test_e2e_jobs.py#L933-L952 should be enough to get the job description and check try value

serhiy-storchaka:
Oh, sorry, I write "There are no tests yet." for the wrong PR. There was a test for this PR. I have added also the CLI test as @asvetlov suggested.

Issue #1529, opened 2020-06-03
Broken dependency with cookiecutter and airflow


yartem:
Problem:
```
$ pip install -U apache-airflow neuro-airflow-plugin
...
$ airflow
Traceback (most recent call last):
  File "/home/ay/.pyenv/versions/3.7.6/bin/airflow", line 25, in <module>
    from airflow.configuration import conf
  File "/home/ay/.pyenv/versions/3.7.6/lib/python3.7/site-packages/airflow/__init__.py", line 97, in <module>
    operators._integrate_plugins()
  File "/home/ay/.pyenv/versions/3.7.6/lib/python3.7/site-packages/airflow/operators/__init__.py", line 104, in _integrate_plugins
    from airflow.plugins_manager import operators_modules, register_inbuilt_operator_links
  File "/home/ay/.pyenv/versions/3.7.6/lib/python3.7/site-packages/airflow/plugins_manager.py", line 194, in <module>
    plugins
  File "/home/ay/.pyenv/versions/3.7.6/lib/python3.7/site-packages/airflow/plugins_manager.py", line 104, in load_entrypoint_plugins
    plugin_obj = entry_point.load()
  File "/home/ay/.local/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2461, in load
    self.require(*args, **kwargs)
  File "/home/ay/.local/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2484, in require
    items = working_set.resolve(reqs, env, installer, extras=self.extras)
  File "/home/ay/.local/lib/python3.7/site-packages/pkg_resources/__init__.py", line 792, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (text-unidecode 1.2 (/home/ay/.pyenv/versions/3.7.6/lib/python3.7/site-packages), Requirement.parse('text-unidecode>=1.3'), {'python-slugify'})
$ pip --version
pip 20.1.1 from /home/ay/.pyenv/versions/3.7.6/lib/python3.7/site-packages/pip (python 3.7)
```
Broken dependencies:
(`->` means `requires`)
[...]

PR #1530, opened 2020-06-03
Fix broken dependency with cookiecutter and airflow

yartem:
Closes https://github.com/neuromation/platform-client-python/issues/1529

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ setup(
         # certifi has no version requirement
         # CLI raises a warning for outdated package instead
         "certifi",
-        "cookiecutter>=0.9.0,<=1.7.1",
+        "cookiecutter>=0.9.0",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -41,7 +41,7 @@
         # certifi has no version requirement
         # CLI raises a warning for outdated package instead
         "certifi",
-        "cookiecutter==1.7.2",
+        "cookiecutter>=0.9.0,<=1.7.1",
```
asvetlov, line 44: Please don't use the upper cap.
Upper limits should be used in `pip install --constraint FILE ...` instruction, not in the library requirements.

yartem: sure. actually, we don't even need the upper constraint.

asvetlov: Great!

Issue #1531, opened 2020-06-05
neuro storage operations implicitly ignore argument part after `?` and `#`


yartem:
Neuro API should raise if input `URL` contains any part that's not used by `neuro storage` operations (parameters folowing after `?`, fragment following after `#`, etc):
```
$ n ls storage:path/?/data.zip
file1 file2
$ n ls storage:path
file1 file2
$ n ls storage:path/data.zip
ERROR: cannot access storage://onprem-poc/tvoinarovskyi/path/data.zip: 404: Not Found
```
Should be fixed here:
https://github.com/neuromation/platform-client-python/blob/18aabf3b8f73ccfdc1a0db7f1a2a4a9284cec086/neuromation/api/storage.py#L93-L101

asvetlov:
Please provide a pull request

yartem:
As we discussed it on standup 11-06-2020, the client should fail with an informative error if the client sees `?` or `#`. The error should say that the user should percent-encode these characters( `?` -> `%3F`, and `#` -> `%23`)

yartem:
And since we will percent-encode these characters so that the whole information will fit into URL's `path`, platform-api needn't to be patched.

asvetlov:
We need a champion for the issue

PR #1538, opened 2020-06-09
Fix neuro-storage commands: local paths to process `?` and `#`

yartem:
Closes https://github.com/neuromation/platform-client-python/issues/1531


yartem:
> Should not normalize_local_path_uri() raise an error instead?

right, as we discussed it on standup 11-06-2020, the client should fail with an informative error if the client sees `?` or `#`

serhiy-storchaka:
We should add checks in two places:
1. For URL arguments in API.
2. When parse string arguments in CLI. Otherwise we would not be able to detect trailing `?` and `#`.

yartem:
Should be patched in a different way:
see https://github.com/neuromation/platform-client-python/issues/1531#issuecomment-642557489 and https://github.com/neuromation/platform-api/issues/1241#issuecomment-642566449

PR #1554, opened 2020-06-16
Secrets

### LINKED ISSUE #1545: Implement secret management commands
**neuro secrets** shows the list of the user's secrets.
```
GET https://dev.neu.ro/api/v1/secrets
```

**neuro secrets add <KEY> <VALUE>** adds a new secrets or replaces an existing one given its key.
```
POST https://dev.neu.ro/api/v1/secrets
{
    "key": "<KEY>",
    "value": "BASE64ENCODED(<VALUE>)"
}
```

**neuro secrets remove <KEY>** removes the secret by its key.
```
DELETE https://dev.neu.ro/api/v1/secrets/<KEY>
```

The backend codebase: https://github.com/neuromation/platform-secrets/blob/ca8e3c3eb9ce8168e09c7ec335b613e64e58e83f/platform_secrets/api.py#L92-L122

asvetlov:
Fixes #1545


neuromation/cli/secrets.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+
+    ret = []
+    async for secret in root.client.secrets.list():
+        ret.append(secret.key)
+        pager_maybe(ret, root.tty, root.terminal_size)
+
+
+@command()
+@argument("key")
```
yartem, line 25: let's call it `secret_name` not `key`

yartem: And also: how about to have API like in curl: 
```
$ neuro secret add secret_name --data pas$w0rd
$ neuro secret add secret_name --data @path/to/file.txt
```

yartem: or the old idea also looks good:
```
$ neuro secret add secret_name pas$w0rd
$ neuro secret add secret_name --from-file path/to/file.txt
```

neuromation/api/secrets.py
```diff
@@ -0,0 +1,42 @@
...
+
+    async def list(self) -> AsyncIterator[Secret]:
+        url = self._config.secrets_url
+        auth = await self._config._api_auth()
+        async with self._core.request("GET", url, auth=auth) as resp:
+            ret = await resp.json()
+            for j in ret:
+                yield Secret(key=j["key"])
+
+    async def add(self, key: str, value: str) -> None:
```
serhiy-storchaka, line 28: Should not `value` be bytes?

neuromation/cli/secrets.py
serhiy-storchaka: I agree that we need a way to add an arbitrary binary value. Docker always read from a file, so for adding a literal value you need to pass it through stdin:
```
$ printf <secret> | docker secret create my_secret -
```
But I think it would be convenient to support passing it as an argument too. Maybe something like:
```
$ neuro secret add secret_name --from-literal pas$w0rd
$ neuro secret add secret_name --from-file path/to/file.txt
```
We can even add an interactive option (in separate issue):
```
$ neuro secret add secret_name
Enter value of secret_name: ***
Repeat value of secret_name: ***
```

neuromation/api/secrets.py
```diff
@@ -0,0 +1,42 @@
...
+from typing import AsyncIterator
+
+from .config import Config
+from .core import _Core
+from .utils import NoPublicConstructor
+
+
+@dataclass(frozen=True)
+class Secret:
+    key: str
```
serhiy-storchaka, line 12: Should not it be a URI?

neuromation/api/secrets.py
asvetlov: Good point. Done

neuromation/api/secrets.py
asvetlov: Why? The server returns just a key.

asvetlov:
I've decided to use `@filename.txt` notation for reading secret value from a file.

asvetlov:
Done!

PR #1556, opened 2020-06-16
Support per-directory .neuroignore files for "neuro cp"

### LINKED ISSUE #1446: Add support of .neuroignore
Currently we can specify filters for file transferring operations either via CLI options (`--exclude` and `--include`) or via the neuro configuration file (see `neuro help user-config`). It may be convenient to specify filters for particular projects, using well known format. There was proposed an idea about using `.gitignore` for this purpose, but it was rejected, because filters for `git` and `neuro` may be different. There was proposed other idea about using `.neuroignore` files with the same format and and semantic as `.gitignore` (see #1182). They will not conflict with `.gitignore`.

serhiy-storchaka:
Closes #1446.

### DIFF
--- a/CHANGELOG.D/1446.feature
+++ b/CHANGELOG.D/1446.feature
@@ -1,0 +1,1 @@
+A `.neuroignore` file specifies files that `neuro cp` should ignore when upload directories to the storage or object storage. Syntax and semantic are similar to `.gitignore` files.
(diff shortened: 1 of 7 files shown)


serhiy-storchaka:
I am not sure. Currently the public API for filtering uses a callable: `Callable[[str], Awaitable[bool]]`. `FileFilter` is just an implementation detail.

It may be wort to add something like `FileFilter` in the standard `fnmatch` module, but it will be not asynchronous.

asvetlov:
Ok, let's keep it a private hidden part

PR #1562, opened 2020-06-17
Fix platform cookies

asvetlov:


asvetlov:
Seems like sticky sessions has weird configuration:
```
sqlite> select distinct(path) from cookie_session;
/api/v1/blob
/
```
I read it as "platform-api and platform-object-storage are configured, platform-storage and platform-registry are not".
/cc: @dalazx

neuromation/api/core.py
```diff
@@ -80,14 +83,28 @@ def __init__(
...
+        storage_url: URL,
+        registry_url: URL,
+        blob_storage_url: URL,
+    ) -> None:
+        for url in (storage_url, registry_url, blob_storage_url):
+            if url is not None:
+                domain = url.host
+                path = url.path
+                if domain is not None:
+                    self._load_cookie(db, domain, path)
```
dalazx, line 98: won't this be overriding the previously set values?

asvetlov: outdated

PR #1583, opened 2020-07-02
Add commands for adding/removing new roles.

### LINKED ISSUE #1582: Add commands for adding/removing roles
We need a command for adding and removing roles.
```
neuro acl add-role mycompany/subdivision
neuro acl remove-role mycompany/subdivision
```

serhiy-storchaka:
Closes #1582.


neuromation/api/users.py
```diff
@@ -88,6 +88,33 @@ async def revoke(self, user: str, uri: URL) -> None:
                 )  # NOQA
         return None
 
+    async def add(self, user: str) -> None:
```
asvetlov, line 91: Is it `user` or `role_name`?

PR #1595, opened 2020-07-09
Enforce ANSI in less command

asvetlov:
Always set "LESS=-R" env variable to fix outputs with scrolling, e.g. "neuro help"

### DIFF
--- a/neuromation/cli/utils.py
+++ b/neuromation/cli/utils.py
@@ -567,6 +567,9 @@ def pager_maybe(
             click.echo(line)
         return
 
+    # Enforce ANSI sequence handling (colors etc.)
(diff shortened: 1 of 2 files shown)


neuromation/cli/utils.py
```diff
@@ -567,6 +567,9 @@ def pager_maybe(
             click.echo(line)
         return
 
+    # Enforce ANSI sequence handling (colors etc.)
+    os.environ["LESS"] = "-R"
```
serhiy-storchaka, line 571: Maybe set LESS only if it was not set? The user can have custom LESS which includes `-R` or `-r`, but also other options.

PR #1616, opened 2020-07-16
Update documentation for upload/download API.

serhiy-storchaka:


asvetlov:
Please merge when green.
Spell checker fails now, I suspect that it doesn't recognize `subdirectories` word in doc files.
The simplest fix is adding this word to `./docs/spelling_wordlist.txt`

PR #1620, opened 2020-07-16
Display job's life-span in "neuro ps"

### LINKED ISSUE #1448: Display job's life-span in neuro ps
Optionally: only if the column is set in `.neuro.toml`

serhiy-storchaka:
Closes #1448.


serhiy-storchaka:
My bad. In CPython the file `topics.py` is automatically generated so I automatically ignore it. :wink:

PR #1622, opened 2020-07-17
Implement images and image tags pagination

### LINKED ISSUE #1606: Implement image pagination
https://github.com/neuromation/platform-registry-api/pull/174

https://docs.docker.com/registry/spec/api/#listing-repositories

serhiy-storchaka:
Closes #1606.


asvetlov (commented):
Everything is done on the server-side, correct?

romasku (approved):
Looks good to me, thx for implementing it on client side.

serhiy-storchaka:
I get HTTP error 504 (Gateway Timeout) when set REPOS_PER_PAGE to too small value (2).

serhiy-storchaka:
Ah, and `test_image_tags` still failed because we only implemented pagination for the list of images, not tags.

PR #1642, opened 2020-07-27
Add pass_config as API method

anayden:
Move `pass_config` functionality from CLI to API, allowing to call it from neuro-extras. Still a draft.


asvetlov:
Sorry, I don't like this idea at all.
Let's discuss the PR offline

PR #1648, opened 2020-07-29
Implement FileStatus.uri property

asvetlov:


neuromation/api/storage.py
```diff
@@ -254,15 +262,16 @@ async def create(self, uri: URL, data: Union[bytes, AsyncIterator[bytes]]) -> No
...
+        url = self._config.storage_url / self._uri_to_path(uri, normalized=True)
         url = url.with_query(op="GETFILESTATUS")
         auth = await self._config._api_auth()
 
         request_time = time.time()
         async with self._core.request("GET", url, auth=auth) as resp:
             self._set_time_diff(request_time, resp)
             res = await resp.json()
-            return _file_status_from_api(res["FileStatus"])
+            return _file_status_from_api(uri.parent, res["FileStatus"])
```
serhiy-storchaka, line 274: What happen when `uri` is `URL("storage://clustername")` or `URL("storage://")`?

asvetlov: good point. I'll handle it in the upcoming commit.

asvetlov: Done.
Surprisingly, the storage server returns the full path for `stat()` call but names only for `ls`.
Seem like a subtle inconsistency for my eyes but I can live with it.

Now `uri.parent` call is not required

CHANGELOG.D/1648.feature
```diff
@@ -0,0 +1 @@
+Implement `FileStatus.uri` property
```
serhiy-storchaka, line 1:
```suggestion
Implement `FileStatus.uri` property.
```

neuromation/api/storage.py
```diff
@@ -693,13 +706,20 @@ def _isrecursive(pattern: str) -> bool:
     return pattern == "**"
 
 
-def _file_status_from_api(values: Dict[str, Any]) -> FileStatus:
+def _file_status_from_api(
+    base_uri: Optional[URL], values: Dict[str, Any]
+) -> FileStatus:
+    if base_uri is None:
+        uri = URL("storage://" + values["path"].lstrip("/"))
```
serhiy-storchaka, line 713: What if the path contains characters `%`, `?` or `#`?

asvetlov: Do you mean `%?#` in cluster or user name?
Is it allowed at all?

serhiy-storchaka: Currently they are not allowed in cluster name and user name. But user can create files with arbitrary names.

asvetlov: Got you.
@serhiy-storchaka please review the updated version

asvetlov:
Thanks @serhiy-storchaka for the very careful review!

PR #1656, opened 2020-08-03
Add find_project_root function

romasku:
This is part of https://github.com/neuromation/neuro-extras/issues/31

### DIFF
--- a/neuromation/api/__init__.py
+++ b/neuromation/api/__init__.py
@@ -145,7 +145,7 @@ async def login(
     *,
     url: URL = DEFAULT_API_URL,
     path: Optional[Path] = None,
-    timeout: aiohttp.ClientTimeout = DEFAULT_TIMEOUT
+    timeout: aiohttp.ClientTimeout = DEFAULT_TIMEOUT,
 ) -> None:
     await Factory(path).login(show_browser_cb, url=url, timeout=timeout)
 
@@ -155,7 +155,7 @@ async def login_with_token(
     *,
     url: URL = DEFAULT_API_URL,
     path: Optional[Path] = None,
-    timeout: aiohttp.ClientTimeout = DEFAULT_TIMEOUT
+    timeout: aiohttp.ClientTimeout = DEFAULT_TIMEOUT,
 ) -> None:
     await Factory(path).login_with_token(token, url=url, timeout=timeout)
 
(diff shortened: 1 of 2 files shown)


neuromation/api/__init__.py
```diff
@@ -145,7 +146,7 @@ async def login(
     *,
     url: URL = DEFAULT_API_URL,
     path: Optional[Path] = None,
-    timeout: aiohttp.ClientTimeout = DEFAULT_TIMEOUT
+    timeout: aiohttp.ClientTimeout = DEFAULT_TIMEOUT,
```
romasku, line 149: I think this changes were introduced by `make fmt`. Should I undo them?

serhiy-storchaka: It looks like a glitch of **black**. If repeated `make fmt` does not change the code you can keep it.

neuromation/api/__init__.py
```diff
@@ -1,3 +1,4 @@
+import pathlib
```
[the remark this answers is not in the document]
romasku: Fixed.

neuromation/api/__init__.py
```diff
@@ -165,10 +165,24 @@ async def login_headless(
...
 
 async def logout(*, path: Optional[Path] = None) -> None:
     await Factory(path).logout()
+
+
+class NoProjectRoot(Exception):
+    pass
+
+
+def find_project_root() -> Path:
```
asvetlov, line 181: Please reuse this function in `load_user_config()` to reduce the code duplication.
Another reason is the future change for search criteria. Perhaps we want to support `.neuro` folder with toml file inside as an option to keep the single space for configs. `neuro-flow` keeps its own scenarios in this way. 
The `.neuro` folder support can wait for a separate PR.

neuromation/api/__init__.py
```diff
@@ -165,10 +165,24 @@ async def login_headless(
...
+    here = Path.cwd()
```
asvetlov, line 182: Please consider optional passing a directory as a starting point. The function signature can look like `def find_project_root(cwd: Optional[Path] = None) -> Path: ...`

neuromation/api/__init__.py
```diff
@@ -165,10 +165,24 @@ async def login_headless(
...
+    while here.parent != here:
+        config = here / ".neuro.toml"
+        if config.exists():
+            return here
+        here = here.parent
+    raise NoProjectRoot(f"Project root is not found for {here}")
```
asvetlov, line 188:
```suggestion
    raise ConfigError(f"Project root is not found for {here}")
```

asvetlov:
Ooops, you've merged the PR just when I was writing the review :)
Please fix notes in the follow-up PR (or convince me that these notes are not relevant).

romasku:
Oh, sorry, I'll do follow-up PR.

asvetlov:
No problem

PR #1657, opened 2020-08-03
Add support of registering plugin config parameters

romasku:
This is part of https://github.com/neuromation/neuro-extras/issues/31

### DIFF
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -87,10 +87,10 @@ Config
       There are two configuration files: **global** and **local**, both are optional
       and can be absent.
 
-      The global file is named `user.toml` and the API search for it in path
-      provided to :class:`Factory` or :func:`get`.
+      The global file is named ``user.toml`` and the API search for it in the path
+      provided to :class:`Factory` or :func:`get` (``$HOME/.neuro/user.toml`` by default).
 
-      The local config file is named `.neuro.toml`, and the API search for this file
+      The local config file is named ``.neuro.toml``, and the API search for this file
       starting from the current folder up to the root directory.
 
       Found local and global configurations are merged.
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -83,8 +83,40 @@ Config
 
    .. comethod:: get_user_config() -> Mapping[str, Any]
 
-      Return user-provided config dictionary, run ``neuro help user-config`` for details
-      about the user configuration files format and location.
+      Return user-provided config dictionary. Config is loaded from config files.
+      There are two configuration files: **global** and **local**, both are optional
+      and can be absent.
+
+      The global file is named `user.toml` and the API search for it in path
+      provided to :class:`Factory` or :func:`get`.
+
+      The local config file is named `.neuro.toml`, and the API search for this file
+      starting from the current folder up to the root directory.
+
+      Found local and global configurations are merged.
+      If a parameter is present are both global and local versions the local parameter
+      take a precedence.
+
+      Configuration files have a TOML format (a stricter version of well-known INI
+      format). See https://en.wikipedia.org/wiki/TOML and
+      https://github.com/toml-lang/toml#toml for the format specification details.
+
+      The API will raise an :class:`ConfigError` if configuration files contains unknown sections or parameters.
+      Note that currently API doesn't use any parameter from user config.
+
+      Known sections: **alias**, **job**, **storage**.
+
+      Section **alias** can have any subsections with any keys.
+
+      Section **job** can have following keys: **ps-format** - string, **life-span** - string.
+
+      Section **storage** can have following keys: **cp-exclude** - list of strings,
+      **cp-exclude-from-files** - list of strings.
+
+      There is a plugin system that allows to register additional config parameters. To
+      define a plugin, add a **neuro_api** entrypoint (check
+      https://packaging.python.org/specifications/entry-points/ for more info about entry points).
+      Entry point should be callable with single argument of type :class:`PluginManager`.
 
       .. versionadded:: 20.01.15
 
--- a/neuromation/api/config.py
+++ b/neuromation/api/config.py
@@ -19,14 +19,11 @@ from yarl import URL
 import neuromation
 
 from .core import _Core
+from .errors import ConfigError
 from .login import AuthTokenClient, _AuthConfig, _AuthToken
 from .plugins import PluginManager
 from .server_cfg import Cluster, Preset, _ServerConfig, get_server_config
 from .utils import NoPublicConstructor, flat
-
-
-class ConfigError(RuntimeError):
-    pass
 
 
 WIN32 = sys.platform == "win32"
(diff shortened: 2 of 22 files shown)


neuromation/api/plugins.py
```diff
@@ -0,0 +1,41 @@
+import numbers
+from typing import Any, Dict, Type, Union
+
+
+class ConfigConflict(Exception):
```
asvetlov, line 5: What is the difference between already existing `ConfigError` and `ConfigConflict` exceptions?

neuromation/api/plugins.py
```diff
@@ -0,0 +1,41 @@
...
+        if section == "alias":
+            raise ConfigConflict("Registering aliases is not supported yet.")
+        if section in self._config_spec and name in self._config_spec[section]:
+            raise ConfigConflict(
+                f"Config parameter {section}.{name} already registered"
+            )
+        self._config_spec.setdefault(section, dict())
+        self._config_spec[section][name] = type
+
+    def register_config_int(self, section: str, name: str) -> None:
```
asvetlov, line 31: I know, I had proposed this name but it looks a little awkward.  Maybe you can suggest better naming?

neuromation/api/config.py
```diff
@@ -578,21 +580,19 @@ def _validate_user_config(
...
-        config, "job", {"ps-format": str, "life-span": str}, filename,
-    )
-    _check_section(
-        config,
-        "storage",
-        {"cp-exclude": (list, str), "cp-exclude-from-files": (list, str)},
-        filename,
-    )
+    plugin_manager = PluginManager()  # Will edit passed dictionary
+    plugin_manager.config.define_str("job", "ps-format")
```
romasku, line 584: We can remove `define_` prefix at all, but I am not sure:
```
plugin_manager.config.str("job", "ps-format")
plugin_manager.config.str("job", "life-span")
plugin_manager.config.str_list("storage", "cp-exclude")
plugin_manager.config.str_list("storage", "cp-exclude-from-files")
```

neuromation/api/plugins.py
```diff
@@ -0,0 +1,69 @@
...
+            Type[numbers.Integral],
+            Type[str],
+            Tuple[Type[List[Any]], Type[bool]],
+            Tuple[Type[List[Any]], Type[str]],
+            Tuple[Type[List[Any]], Type[numbers.Real]],
+            Tuple[Type[List[Any]], Type[numbers.Integral]],
+        ],
+    ) -> None:
+        # Avoid circular import config.py <-> plugins.py
+        from .config import ConfigError
```
romasku, line 27: We can move Exceptions to special file (like `exceptions.py`) to avoid problems like this, but it disagrees with "feature oriented" file structure.

neuromation/api/config.py
asvetlov: Let's keep `define_`

neuromation/api/config.py
```diff
@@ -578,21 +580,19 @@ def _validate_user_config(
...
-    _check_section(
-        config, "job", {"ps-format": str, "life-span": str}, filename,
-    )
-    _check_section(
-        config,
-        "storage",
-        {"cp-exclude": (list, str), "cp-exclude-from-files": (list, str)},
-        filename,
-    )
+    plugin_manager = PluginManager()  # Will edit passed dictionary
```
asvetlov, line 583:
```suggestion
    plugin_manager = PluginManager()
```

neuromation/api/plugins.py
asvetlov: Doesn't matter really: the only public name should be `neuromation.api` anyway.
People should avoid importing nested modules.

A separate file for exceptions looks good in this case.

neuromation/api/plugins.py
```diff
@@ -0,0 +1,69 @@
...
+    def define_float_list(self, section: str, name: str) -> None:
+        self._define_param(section, name, (list, numbers.Real))
+
+
+class PluginManager:
+
+    config: ConfigBuilder
+
+    def __init__(self) -> None:
+        self.config = ConfigBuilder()
```
asvetlov, line 69: Please use `_config` attr and `config` readonly property for the sake of uniformness.

neuromation/api/plugins.py
```diff
@@ -0,0 +1,69 @@
...
+        from .config import ConfigError
+
+        if section == "alias":
+            raise ConfigError("Registering aliases is not supported yet.")
+        if section in self._config_spec and name in self._config_spec[section]:
+            raise ConfigError(f"Config parameter {section}.{name} already registered")
+        self._config_spec.setdefault(section, dict())
+        self._config_spec[section][name] = type
+
+    def get_spec(self) -> Mapping[str, Any]:
```
asvetlov, line 36: I doubt if the method should be public.
Maybe we will change the implementation completely.
`define_int()` and family has a great chance to survive as-as; `get_spec()` returns internal implementation details and very likely will be changed.

Also, please think about the documentation. How to describe the returned value to the end-user? Why the user need this method at all?

neuromation/api/plugins.py
romasku: I was more thinking about us, not public api users. I mean that it isn't very comfortable to have exception outside of the only file it is used. Also we cannot inforce this rule

asvetlov: I don't insist, at least for now.

docs/config_reference.rst
```diff
@@ -83,8 +83,40 @@ Config
...
    .. comethod:: get_user_config() -> Mapping[str, Any]
 
-      Return user-provided config dictionary, run ``neuro help user-config`` for details
-      about the user configuration files format and location.
+      Return user-provided config dictionary. Config is loaded from config files.
+      There are two configuration files: **global** and **local**, both are optional
+      and can be absent.
+
+      The global file is named `user.toml` and the API search for it in path
+      provided to :class:`Factory` or :func:`get`.
```
asvetlov, line 91:
```suggestion
      provided to :class:`Factory` or :func:`get` (``$HOME/.neuro/user.toml`` by default).
```

docs/config_reference.rst
```diff
@@ -83,8 +83,40 @@ Config
...
+
+      The local config file is named `.neuro.toml`, and the API search for this file
```
asvetlov, line 93:
```suggestion
      The local config file is named ``.neuro.toml``, and the API search for this file
```

docs/config_reference.rst
```diff
@@ -83,8 +83,40 @@ Config
 
    .. comethod:: get_user_config() -> Mapping[str, Any]
 
-      Return user-provided config dictionary, run ``neuro help user-config`` for details
-      about the user configuration files format and location.
+      Return user-provided config dictionary. Config is loaded from config files.
+      There are two configuration files: **global** and **local**, both are optional
+      and can be absent.
+
+      The global file is named `user.toml` and the API search for it in path
```
asvetlov, line 90:
```suggestion
      The global file is named ``user.toml`` and the API search for it in the path
```

PR #1660, opened 2020-08-04
Add find_project_root (follow-up)

romasku:
This is follow-up PR to https://github.com/neuromation/platform-client-python/pull/1656

### DIFF
--- a/neuromation/api/utils.py
+++ b/neuromation/api/utils.py
@@ -1,6 +1,7 @@
 import asyncio
 import logging
 import sys
+from pathlib import Path
 from types import TracebackType
 from typing import (
     Any,
@@ -17,6 +18,8 @@ from typing import (
 )
 
 import aiohttp
+
+from neuromation.api.errors import ConfigError
 
 
 if sys.version_info >= (3, 7):  # pragma: no cover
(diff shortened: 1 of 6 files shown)


neuromation/api/utils.py
```diff
@@ -112,3 +113,15 @@ async def retry() -> AsyncIterator[None]:
...
+
+
+def find_project_root(cwd: Optional[Path] = None) -> Path:
+    here = cwd if cwd is not None else Path.cwd()
+    while here.parent != here:
+        config = here / ".neuro.toml"
+        if config.exists():
+            return here
+        here = here.parent
+    from .config import ConfigError
```
romasku, line 125: We cannot avoid circular import here: `config.py` uses `find_project_root`, but it requires `ConfigError` to be imported.
Also I am still not sure that `ConfigError` is correct exeption here - I though it means "you config file is bad, delete it and make new one`, and here we just unable to find it at all.

romasku: Will be able to fix this when after https://github.com/neuromation/platform-client-python/pull/1657 is merged.

asvetlov: ok, let's keep the circular import.
`ConfigError` is probably good enough.
We can play building the comprehensive exceptions hierarchy game, but I prefer to do it on demand only.

neuromation/api/utils.py
```diff
@@ -112,3 +115,13 @@ async def retry() -> AsyncIterator[None]:
 
 def flat(sql: str) -> str:
     return " ".join(line.strip() for line in sql.splitlines() if line.strip())
+
+
+def find_project_root(cwd: Optional[Path] = None) -> Path:
```
serhiy-storchaka, line 120: I am not sure `cwd` is good name.

romasku: I agree that `path` is more appropriate here, thanks for suggestion.

asvetlov:
Please merge

PR #1667, opened 2020-08-07
Feature/replace queued progress

romasku:
Refactors out `QueuedProgress` to allow to use idea "store calls into `asyncio.Queue`" not only for `AbstractFileProgress` descendants.
Required for https://github.com/neuromation/platform-client-python/pull/1664

### DIFF
--- a/neuromation/api/utils.py
+++ b/neuromation/api/utils.py
@@ -17,7 +17,6 @@ from typing import (
     Tuple,
     Type,
     TypeVar,
-    cast,
 )
 
 import aiohttp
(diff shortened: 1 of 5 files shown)


neuromation/api/utils.py
```diff
@@ -127,3 +129,54 @@ def find_project_root(path: Optional[Path] = None) -> Path:
...
+            # Force TypeError
+            real_method()  # NOQA
+        await queue.put(partial(real_method, *args, **kwargs))
+
+    async def noop(*args: Any, **kwargs: Any) -> None:
+        pass
+
+    class Proxy:
+        def __getattr__(self, item: str) -> Callable[[Any], Coroutine[Any, Any, None]]:
+            if any_obj is None and allow_any_for_none:
```
serhiy-storchaka, line 177: This condition does not change its value, so you can check it at the time of creation the `Proxy` class.

neuromation/api/utils.py
```diff
@@ -127,3 +129,54 @@ def find_project_root(path: Optional[Path] = None) -> Path:
...
+    raise any AttributeErrors and just absorb all cals silently.
+    """
+    queue: "asyncio.Queue[QueuedCall]" = asyncio.Queue()
+
+    async def add_to_queue(
+        real_method: Callable[..., None], *args: Any, **kwargs: Any
+    ) -> None:
+        if not callable(real_method):
+            # Force TypeError
+            real_method()  # NOQA
```
serhiy-storchaka, line 169: Does not it add some overhead at run time?

neuromation/api/utils.py
```diff
@@ -127,3 +129,54 @@ def find_project_root(path: Optional[Path] = None) -> Path:
...
+
+    async def noop(*args: Any, **kwargs: Any) -> None:
+        pass
+
+    class Proxy:
+        def __getattr__(self, item: str) -> Callable[[Any], Coroutine[Any, Any, None]]:
+            if any_obj is None and allow_any_for_none:
+                return noop
+            real_method = getattr(any_obj, item)
+            return partial(add_to_queue, real_method)
```
serhiy-storchaka, line 180: Hint: you can speed up the method lookup by caching it as the attribute of the proxy object.
```python
method = partial(add_to_queue, real_method)
setattr(self, item, method)
return method
```

neuromation/api/utils.py
```diff
@@ -127,3 +129,54 @@ def find_project_root(path: Optional[Path] = None) -> Path:
...
+        if not callable(real_method):
+            # Force TypeError
+            real_method()  # NOQA
+        await queue.put(partial(real_method, *args, **kwargs))
+
+    async def noop(*args: Any, **kwargs: Any) -> None:
+        pass
+
+    class Proxy:
+        def __getattr__(self, item: str) -> Callable[[Any], Coroutine[Any, Any, None]]:
```
serhiy-storchaka, line 176: Is not `name` a better parameter name for the name of attribute?

PR #1697, opened 2020-08-20
Handle errors messages from delete response stream

romasku:
Adds support of https://github.com/neuromation/platform-storage-api/pull/239

### DIFF
--- a/neuromation/api/storage.py
+++ b/neuromation/api/storage.py
@@ -20,6 +20,7 @@ from typing import (
     Callable,
     Dict,
     Iterable,
+    Mapping,
     Optional,
     Tuple,
     Union,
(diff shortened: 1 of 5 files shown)


neuromation/api/storage.py
```diff
@@ -338,11 +338,18 @@ async def _rm(
         ) as resp:
             if resp.headers.get("Content-Type", "").startswith("application/x-ndjson"):
                 async for line in resp.content:
-                    remove_listing = json.loads(line)
+                    server_message = json.loads(line)
+                    if "error" in server_message:
+                        err_text = server_message["error"]
+                        os_errno = server_message.get("errno", None)
+                        if os_errno is not None:
+                            os_errno = errno.__dict__.get(os_errno, os_errno)
```
asvetlov, line 346: I'm missed. What is `errno` type here?

romasku: `errno` here is `module` instance (it comes from `import errno`). This is code is almost same as in `_Core.request`.

asvetlov: Got you

asvetlov (approved):

serhiy-storchaka (commented):
Could you please add similar code for other response stream (`storage ls`, `job ls`)?

PR #1698, opened 2020-08-20
Colorize neuro ps output

asvetlov:
The change makes the output more readable.

### DIFF
--- a/neuromation/cli/formatters/jobs.py
+++ b/neuromation/cli/formatters/jobs.py
@@ -288,10 +288,10 @@ class TabularJobRow:
             when = job.history.finished_at
         assert when is not None
         return cls(
-            id=job.id,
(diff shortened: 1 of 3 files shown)


neuromation/cli/formatters/jobs.py
```diff
@@ -288,10 +288,10 @@ def from_job(
             when = job.history.finished_at
         assert when is not None
         return cls(
-            id=job.id,
+            id=style(job.id, bold=True),
```
serhiy-storchaka, line 291: You can use `bold()` defined at the top of this file.

asvetlov: thanks!

serhiy-storchaka (approved):

PR #1707, opened 2020-08-31
Refactor secrets parsing

anayden:
Extract code handling secrets parsing from CLI to Parser in order to reuse it in neuro-extras.

### DIFF
--- a/CHANGELOG.D/1707.feature
+++ b/CHANGELOG.D/1707.feature
@@ -1,0 +1,1 @@
+Deprecate `--volume=ALL` option in CLI and move volume and env variable parsing from CLI module to the `Parser` class in API
--- a/CHANGELOG.D/1707.feature
+++ b/CHANGELOG.D/1707.feature
@@ -1,1 +1,1 @@
-Deprecate `--volume=ALL` option in CLI and move volume and env variable parsing from CLI module to the `Parser` class in API
+Remove `--volume=ALL` option from CLI and move volume and env variable parsing from CLI module to the `Parser` class in API
--- a/docs/parse_reference.rst
+++ b/docs/parse_reference.rst
@@ -52,7 +52,7 @@ Parser
 
   .. _DockerHub: https://hub.docker.com
 
-   .. method:: env(env: Sequence[str], env_file: Sequence[str] = ()) -> Tuple[Dict[str, str], Mapping[str, URL]]
+   .. method:: env(env: Sequence[str], env_file: Sequence[str] = ()) -> Tuple[Dict[str, str], Dict[str, URL]]
 
       Parse a sequence of *env* variables and a sequence of *env_file* file names into a
       tuple of two mappings - first one for all environment variables and second for
--- a/docs/parse_reference.rst
+++ b/docs/parse_reference.rst
@@ -58,8 +58,6 @@ Parser
       tuple of two mappings - first one for all environment variables and second for
       environment variables using secrets.
 
-      :raise: :exc:`ValueError` if caller attempts to redefine a system-reserved environment variable
-
    .. method:: volumes(volume: Sequence[str]) -> Tuple[Set[Volume], Set[SecretFile]]
 
       Parse a sequence of volume definition into a tuple of two mappings - first one for
--- a/docs/parse_reference.rst
+++ b/docs/parse_reference.rst
@@ -50,4 +50,20 @@ Parser
 
       :raise: :exc:`ValueError` if *image* has invalid format.
 
+   .. method:: env(env: Sequence[str], env_file: Sequence[str] = ()) -> Tuple[Dict[str, str], Mapping[str, URL]]
+
+      Parse a sequence of *env* variables and a sequence of *env_file* file names into a
+      tuple of two mappings - first one for all environment variables and second for
+      environment variables using secrets.
+
+      :raise: :exc:`ValueError` if caller attempts to redefine a system-reserved env var
+
+   .. method:: volumes(volume: Sequence[str]) -> Tuple[Set[Volume], Set[SecretFile]]
+
+      Parse a sequence of volume definition into a tuple of two mappings - first one for
+      all regular volumes and second one for volumes using secrets.
+
+      :raise: :exc:`ValueError` if `volumes` sequence contains `ALL` or `HOME` values,
+      which are deprecated.
+
 .. _DockerHub: https://hub.docker.com
--- a/docs/parse_reference.rst
+++ b/docs/parse_reference.rst
@@ -58,7 +58,7 @@ Parser
       tuple of two mappings - first one for all environment variables and second for
       environment variables using secrets.
 
-      :raise: :exc:`ValueError` if caller attempts to redefine a system-reserved env var
+      :raise: :exc:`ValueError` if caller attempts to redefine a system-reserved environment variable
 
    .. method:: volumes(volume: Sequence[str]) -> Tuple[Set[Volume], Set[SecretFile]]
 
--- a/docs/parse_reference.rst
+++ b/docs/parse_reference.rst
@@ -65,6 +65,3 @@ Parser
       Parse a sequence of volume definition into a tuple of two mappings - first one for
       all regular volumes and second one for volumes using secrets.
 
-      :raise: :exc:`ValueError` if `volumes` sequence contains `ALL` or `HOME` values, which are deprecated.
-
-
--- a/docs/parse_reference.rst
+++ b/docs/parse_reference.rst
@@ -58,7 +58,7 @@ Parser
       tuple of two mappings - first one for all environment variables and second for
       environment variables using secrets.
 
-   .. method:: volumes(volume: Sequence[str]) -> Tuple[Set[Volume], Set[SecretFile]]
+   .. method:: volumes(volume: Sequence[str]) -> Tuple[List[Volume], List[SecretFile]]
 
       Parse a sequence of volume definition into a tuple of two mappings - first one for
       all regular volumes and second one for volumes using secrets.
--- a/docs/parse_reference.rst
+++ b/docs/parse_reference.rst
@@ -50,6 +50,8 @@ Parser
 
       :raise: :exc:`ValueError` if *image* has invalid format.
 
+  .. _DockerHub: https://hub.docker.com
+
    .. method:: env(env: Sequence[str], env_file: Sequence[str] = ()) -> Tuple[Dict[str, str], Mapping[str, URL]]
 
       Parse a sequence of *env* variables and a sequence of *env_file* file names into a
@@ -63,7 +65,6 @@ Parser
       Parse a sequence of volume definition into a tuple of two mappings - first one for
       all regular volumes and second one for volumes using secrets.
 
-      :raise: :exc:`ValueError` if `volumes` sequence contains `ALL` or `HOME` values,
-      which are deprecated.
+      :raise: :exc:`ValueError` if `volumes` sequence contains `ALL` or `HOME` values, which are deprecated.
 
-.. _DockerHub: https://hub.docker.com
+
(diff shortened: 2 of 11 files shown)


neuromation/api/parser.py
```diff
@@ -1,13 +1,27 @@
+import os
 from dataclasses import dataclass
+from typing import Dict, Iterator, Optional, Sequence, Set
 
+import click
```
asvetlov, line 5: `click` should not be used in API

neuromation/api/parser.py
```diff
@@ -39,6 +53,25 @@ def volume(self, volume: str) -> Volume:
             storage_uri=storage_uri, container_path=container_path, read_only=read_only
         )
 
+    def build_secret_files(self, input_volumes: Set[str]) -> Set[SecretFile]:
```
asvetlov, line 56: Here and below: the api call looks like `client.parse.build_secret_files(...)`. 
`parse.build_xxx()` looks weird, please choose another name for the method.

asvetlov: Another problem is the method signature: it returns a set of secret files that should be extracted from `input_volumes` at the next step. 
It is not very handy and hard to explain (try to write lucid API docs as the exercise).
I can suggest another API: `client.parse.volumes()` that returns a named tuple of regular and secret volume sequences, e.g. `volumes, secret_files = client.parse.volumes(inp_vols)`
What do you think?

neuromation/api/parser.py
```diff
@@ -39,6 +53,25 @@ def volume(self, volume: str) -> Volume:
...
+        for volume in input_volumes:
+            parts = volume.split(":")
+            if len(parts) != 3:
+                raise ValueError(f"Invalid secret file specification '{volume}'")
+            container_path = parts.pop()
+            secret_uri = self.parse_secret_resource(":".join(parts))
+            secret_files.add(SecretFile(secret_uri, container_path))
+        return secret_files
+
+    def parse_secret_resource(self, uri: str) -> URL:
```
asvetlov, line 67: Should the method be public or private?
What is the usage outside of the `Parser` class?

neuromation/api/parser.py
```diff
@@ -64,3 +97,39 @@ def _remote_to_local_image(self, image: RemoteImage) -> LocalImage:
             self._config.username, self._config.cluster_name, self._config.registry_url
         )
         return parser.convert_to_local_image(image)
+
+    def build_env(self, env: Sequence[str], env_file: Optional[str]) -> Dict[str, str]:
```
asvetlov, line 101: The same thing: the naming and functionality.
Instead of *build/split* pair, we can have `client.parse.env(env, env_file)` method that returns a pair of `Mapping[str, str]` for regular env-vars and `Mapping[str, URL]` for secret envs.

asvetlov (changes requested):
Sorry, I have a different vision of API.
Let's discuss it.

Also, the parser is a part of public API.
New methods should be documented.

neuromation/api/parser.py
anayden: I really like this clean approach, but parsing (non-secret) volumes with the `--ALL` option, which is implemented in `_build_volumes` seems to be fairly complicated, as it queries user's ACL. Do you think `Parser` should do that or how would you address this particular option for volumes?

neuromation/api/parser.py
anayden: 👍

neuromation/api/parser.py
asvetlov: We can handle `"ALL"` on the CLI side but parse all other values in API parser.

neuromation/api/parser.py
serhiy-storchaka: Docker supports multiple env files (https://docs.docker.com/compose/compose-file/#env_file), so it may be better to accept a sequence of strings.

serhiy-storchaka: #1711

asvetlov: @serhiy-storchaka good idea

anayden:
@asvetlov @serhiy-storchaka I've implemented changes you asked for. There's a flacky e2e test on job save, but other than that the code is complete in my opinion. Would you mind to take a look again?

neuromation/api/parser.py
serhiy-storchaka: I agree with @asvetlov that it likely should be private method. If it should be public, rename it to `secret_resource()` or `secret_uri()` or something like.

asvetlov:
Sure! Please let me review tomorrow morning.

neuromation/api/parser.py
```diff
@@ -64,3 +108,61 @@ def _remote_to_local_image(self, image: RemoteImage) -> LocalImage:
...
+def _read_lines(env_file: str) -> Iterator[str]:
+    with open(env_file, encoding="utf-8-sig") as ef:
+        lines = ef.read().splitlines()
+    for line in lines:
+        line = line.lstrip()
+        if line and not line.startswith("#"):
+            yield line
+
+
+def _get_neuro_mountpoint(username: str) -> str:
```
serhiy-storchaka: Not used.

neuromation/api/parser.py
```diff
@@ -64,3 +108,61 @@ def _remote_to_local_image(self, image: RemoteImage) -> LocalImage:
             self._config.username, self._config.cluster_name, self._config.registry_url
         )
         return parser.convert_to_local_image(image)
+
+    def env(
+        self, env: Sequence[str], env_file: Sequence[str] = ()
+    ) -> Tuple[Dict[str, str], Mapping[str, URL]]:
```
serhiy-storchaka, line 114: Why `Tuple[Dict[str, str], Mapping[str, URL]]`, but not `Tuple[Dict[str, str], Dict[str, URL]]` or `Tuple[Mapping[str, str], Mapping[str, URL]]`?

neuromation/api/parser.py
```diff
@@ -64,3 +108,61 @@ def _remote_to_local_image(self, image: RemoteImage) -> LocalImage:
...
+            splitted = line.split("=", 1)
+            name = splitted[0]
+            if len(splitted) == 1:
+                val = os.environ.get(splitted[0], "")
+            else:
+                val = splitted[1]
+            if name in RESERVED_ENV_VARS:
+                raise ValueError(
+                    f"Unable to re-define system-reserved environment variable: {name}"
+                )
```
serhiy-storchaka, line 138: Why forbid this?

neuromation/api/parser.py
```diff
@@ -16,8 +32,9 @@ class Volume:
 
 
 class Parser(metaclass=NoPublicConstructor):
-    def __init__(self, config: Config) -> None:
+    def __init__(self, config: Config, users: Users) -> None:
         self._config = config
+        self._users = users
```
serhiy-storchaka, line 37: Seems it is not used.

CHANGELOG.D/1707.feature
```diff
@@ -0,0 +1 @@
+Deprecate `--volume=ALL` option in CLI and move volume and env variable parsing from CLI module to the `Parser` class in API
```
serhiy-storchaka, line 1: It is not deprecated. It is now just not supported.

docs/parse_reference.rst
```diff
@@ -50,4 +50,21 @@ Parser
...
+      environment variables using secrets.
+
+      :raise: :exc:`ValueError` if caller attempts to redefine a system-reserved env var
+
+   .. method:: volumes(volume: Sequence[str]) -> Tuple[Set[Volume], Set[SecretFile]]
+
+      Parse a sequence of volume definition into a tuple of two mappings - first one for
+      all regular volumes and second one for volumes using secrets.
+
+      :raise: :exc:`ValueError` if `volumes` sequence contains `ALL` or `HOME` values, which are deprecated.
```
serhiy-storchaka, line 68: I think it is not needed. First, they are not deprecated, they are errors. Second, a ValueError is raised for any other invalid volume specification.

neuromation/api/parser.py
anayden: This has been the case historically. I'm afraid I don't have context. Perhaps @yartem can help?

serhiy-storchaka: It has been to prevent conflicts with `--volume=ALL` and `--volume=HOME`. But since these features are no longer supported these environment variables are no longer set by client.

anayden: Ah, I see. Removed it, as well as corresponding docs and tests.

neuromation/api/parser.py
```diff
@@ -1,13 +1,24 @@
...
 from yarl import URL
 
 from .config import Config
 from .parsing_utils import LocalImage, RemoteImage, TagOption, _ImageNameParser
-from .url_utils import normalize_storage_path_uri
+from .url_utils import normalize_storage_path_uri, uri_from_cli
 from .utils import NoPublicConstructor
 
 
+ROOT_MOUNTPOINT = "/var/neuro"
```
serhiy-storchaka, line 13: Seems it is not used anymore too.

neuromation/api/parser.py
```diff
@@ -64,3 +102,53 @@ def _remote_to_local_image(self, image: RemoteImage) -> LocalImage:
...
+
+    def _extract_secret_env(self, env_dict: Dict[str, str]) -> Dict[str, URL]:
+        secret_env_dict = {}
+        for name, val in env_dict.copy().items():
+            if val.startswith("secret:"):
+                secret_env_dict[name] = self._parse_secret_resource(val)
+                del env_dict[name]
+        return secret_env_dict
+
+    def volumes(self, volume: Sequence[str]) -> Tuple[Set[Volume], Set[SecretFile]]:
```
asvetlov, line 140: Please replace `Set[...]` with `Sequence[...]` in the signature and return a pair of lists actually.
The reason is: `api.Container` has sequence types for `volumes` and `secret_files`, let's keep the consistency.

neuromation/api/parser.py
asvetlov: Agree, please remove

asvetlov (commented):
Almost perfect, please fix two minor comments.

neuromation/api/parser.py
asvetlov: `List[...]` is even more appropriate data type, as you use `Dict` for envs

anayden: Done, but I had to add `volumes = list(volumes)` in cli code in order to make the var mutable.

docs/parse_reference.rst
```diff
@@ -50,4 +50,16 @@ Parser
...
-.. _DockerHub: https://hub.docker.com
+  .. _DockerHub: https://hub.docker.com
+
+   .. method:: env(env: Sequence[str], env_file: Sequence[str] = ()) -> Tuple[Dict[str, str], Dict[str, URL]]
+
+      Parse a sequence of *env* variables and a sequence of *env_file* file names into a
+      tuple of two mappings - first one for all environment variables and second for
+      environment variables using secrets.
+
+   .. method:: volumes(volume: Sequence[str]) -> Tuple[Set[Volume], Set[SecretFile]]
```
serhiy-storchaka, line 61: Update annotation.

neuromation/api/parser.py
```diff
@@ -64,3 +99,55 @@ def _remote_to_local_image(self, image: RemoteImage) -> LocalImage:
...
+        secret_env_dict = {}
+        for name, val in env_dict.copy().items():
+            if val.startswith("secret:"):
+                secret_env_dict[name] = self._parse_secret_resource(val)
+                del env_dict[name]
+        return secret_env_dict
+
+    def volumes(
+        self, volume: Sequence[str]
+    ) -> Tuple[Sequence[Volume], Sequence[SecretFile]]:
```
serhiy-storchaka, line 139: I concur with @asvetlov that we can return `List`s.

serhiy-storchaka (approved):
:+1:

asvetlov (approved):
lgtm

PR #1709, opened 2020-08-31
Add support for filtering ACL lists by the URI prefix

### LINKED ISSUE #1708: Add support for filtering by the URI prefix in "neuro acl list"
Currently you can request the list filtered by scheme by specifying the `--scheme` option:
```
neuro acl list --scheme=storage
neuro acl list --shared --scheme=storage
```
It is possible to generalize this by filtering by the URI prefix instead of just scheme. The above command would be equivalent to the following commands:
```
neuro acl list storage:
neuro acl list --shared storage:
```
But you can specify also larger prefix, e.g. `storage://mycluster/myuser/mydir`.

Unfortunately we cannot use "short URIs" (e.g. `scheme:path` and `scheme:/user/path`) because we need a way to specify "all URIs with specified scheme", but short URI `storage:` would expand to `storage://mycluster/myuser`. The argument of this command should be an absolute URI.

Suppoort of short URI can be added if we find a way to specify "all URIs with specified scheme" different from `scheme:`.

serhiy-storchaka:
Closes #1708.

### DIFF
--- a/CHANGELOG.D/1708.feature
+++ b/CHANGELOG.D/1708.feature
@@ -1,0 +1,1 @@
+Command `neuro acl list` accepts now a URI prefix for filtering. API functions `users.get_acl()` and `users.get_shares()` accept now the *uri* argument.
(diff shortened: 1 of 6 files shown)


asvetlov (approved):
lgtm!

PR #1711, opened 2020-09-01
Support multiple --env-file options in "neuro run"

### LINKED ISSUE #1710: Support multiple --env-file options in "neuro run"
Docker supports multiple env files (https://docs.docker.com/compose/compose-file/#env_file), so it may be worth to support multiple `--env-file` options in `neuro run`

serhiy-storchaka:
Closes #1710.


asvetlov (approved):
LGTM, but please add a change note file before merging

anayden (approved):
Nice!

PR #1716, opened 2020-09-03
Add disks management commands

romasku:
Part of https://github.com/neuromation/platform-client-python/issues/1715

### DIFF
--- a/neuromation/api/disks.py
+++ b/neuromation/api/disks.py
@@ -1,0 +1,75 @@
+from dataclasses import dataclass
+from enum import Enum
+from typing import Any, AsyncIterator, Mapping
+
+from yarl import URL
+
+from .config import Config
+from .core import _Core
+from .utils import NoPublicConstructor
+
+
+@dataclass(frozen=True)
+class Disk:
+    id: str
+    storage: int  # In bytes
+    owner: str
+    status: "Disk.Status"
+    _cluster: str
+
+    @property
+    def uri(self) -> URL:
+        return URL(f"disk://{self._cluster}/{self.owner}/{self.id}")
+
+    class Status(Enum):
+        PENDING = "Pending"
+        READY = "Ready"
+        BROKEN = "Broken"
+
+
+class Disks(metaclass=NoPublicConstructor):
+    def __init__(self, core: _Core, config: Config) -> None:
+        self._core = core
+        self._config = config
+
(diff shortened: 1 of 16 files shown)


neuromation/cli/formatters/disks.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def _disk_to_table_row(self, disk: Disk) -> Sequence[str]:
+        if disk.storage >= 1024 ** 3:
+            storage_str = f"{disk.storage / (1024 ** 3):.2f}G"
+        elif disk.storage >= 1024 ** 2:
+            storage_str = f"{disk.storage / (1024 ** 2):.2f}M"
+        elif disk.storage >= 1024:
+            storage_str = f"{disk.storage / 1024:.2f}K"
+        else:
+            storage_str = str(disk.storage)
```
serhiy-storchaka, line 29: Can it use `cli.utils.format_size()`?

romasku: Yes, sure, thanks for pointing out.

neuromation/api/disks.py
```diff
@@ -0,0 +1,75 @@
...
+from .utils import NoPublicConstructor
+
+
+@dataclass(frozen=True)
+class Disk:
+    id: str
+    storage: int  # In bytes
+    owner: str
+    status: "Disk.Status"
+    _cluster: str
```
serhiy-storchaka, line 18: Why is it underscored?

romasku: At first I thought that this is some kind of internal data (as it isn't returned from server), but now I see no reason to underscore this field.

neuromation/api/disks.py
```diff
@@ -0,0 +1,73 @@
...
+from .utils import NoPublicConstructor
+
+
+@dataclass(frozen=True)
+class Disk:
+    id: str
+    storage: int  # In bytes
+    owner: str
+    status: "Disk.Status"
+    cluster: str
```
serhiy-storchaka, line 18: In all other classes (`_ConfigData`, `JobDescription`, `RemoteImage`, `_QuotaInfo`) we use attributes named `cluster_name`. Name `cluster` is never used. It is rather by accident, but it would be better to continue the tradition.

serhiy-storchaka (approved):
:+1:

PR #1721, opened 2020-09-04
Add support of disk_volumes in neuro run

romasku:
Part of https://github.com/neuromation/platform-client-python/issues/1715

### DIFF
--- a/neuromation/api/jobs.py
+++ b/neuromation/api/jobs.py
@@ -41,10 +41,14 @@ from .images import (
     _raise_on_error_chunk,
     _try_parse_image_progress_step,
 )
-from .parser import Parser, SecretFile, Volume
+from .parser import DiskVolume, Parser, SecretFile, Volume
 from .parsing_utils import LocalImage, RemoteImage, _as_repo_str, _is_in_neuro_registry
 from .server_cfg import Preset
-from .url_utils import normalize_secret_uri, normalize_storage_path_uri
+from .url_utils import (
+    normalize_disk_uri,
+    normalize_secret_uri,
+    normalize_storage_path_uri,
+)
 from .utils import NoPublicConstructor, asynccontextmanager
 
 
@@ -102,6 +106,7 @@ class Container:
     volumes: Sequence[Volume] = field(default_factory=list)
     secret_env: Mapping[str, URL] = field(default_factory=dict)
     secret_files: Sequence[SecretFile] = field(default_factory=list)
+    disk_volumes: Sequence[DiskVolume] = field(default_factory=list)
     tty: bool = False
 
 
@@ -680,6 +685,7 @@ def _container_from_api(data: Dict[str, Any], parse: Parser) -> Container:
         volumes=[_volume_from_api(v) for v in data.get("volumes", [])],
         secret_env={name: URL(val) for name, val in data.get("secret_env", {}).items()},
         secret_files=[_secret_file_from_api(v) for v in data.get("secret_volumes", [])],
+        disk_volumes=[_disk_volume_from_api(v) for v in data.get("disk_volumes", [])],
         tty=data.get("tty", False),
     )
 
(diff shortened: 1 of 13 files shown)


romasku:
E2E test requires https://github.com/neuromation/platform-client-python/pull/1716 to be merged.

neuromation/api/jobs.py
```diff
@@ -41,10 +41,14 @@
     _raise_on_error_chunk,
     _try_parse_image_progress_step,
 )
-from .parser import Parser, SecretFile, Volume
+from .parser import DiskVolume, Parser, SecretFile, Volume
```
asvetlov, line 44: Please replace `DiskVolume` -> `Disk`, `disk_volumes_` -> `disks` etc. here and below.
Shorter is better.

neuromation/api/parser.py
```diff
@@ -134,12 +162,16 @@ def _extract_secret_env(self, env_dict: Dict[str, str]) -> Dict[str, URL]:
                 del env_dict[name]
         return secret_env_dict
 
-    def volumes(self, volume: Sequence[str]) -> Tuple[List[Volume], List[SecretFile]]:
+    def volumes(
+        self, volume: Sequence[str]
+    ) -> Tuple[List[Volume], List[SecretFile], List[DiskVolume]]:
```
asvetlov, line 167: Breaking change, `volumes, secrets = client.parse.volumes(...)` doesn't work anymore.

To fix this problem once and forever please return a dataclass with `volumes`, `secret_files`, and `disks` field instead of a tuple.
The API was added this week, better to fix it quickly.

asvetlov: Please apply the same change to `env` method. 
Please Feel free to rename `env` -> `envs` to mimic `volumes`.
The doc should reflect changes.

asvetlov (commented):
looks good except a few minor design notes.

asvetlov:
BTW, a changelog file is required

neuromation/api/jobs.py
romasku: I'm not sure how to do this properly. In another PR(https://github.com/neuromation/platform-client-python/pull/1716) I  add another `Disk` dataclass, so I cannot rename this one. Also in my mind the disk and disk volume is two different concepts (same way as **Secret** and **SecretFile**):
- **Disk** is a storage device, it has some capacity, status, owner, and lifespan in the future. It can exist without any container.
- **DiskVolume** is just specification on how to mount some **Disk** to some container.

On the other side, I am not sure that naming is correct. We can use something like **DiskMountSpec**, but it even longer.

Also currently `DiskVolume` and `Volume` dataclasses have no real difference. We can reuse `Volume` for disk mounts, and make API code separate them just before sending it to the server. For user code, there will be no difference between mounting `storage://` or `disk://`. But this is somehow implicit logic as for me and I prefer to avoid such a solution.

asvetlov: ok, got it

neuromation/api/parser.py
```diff
@@ -23,26 +23,50 @@ class Volume:
...
+@dataclass(frozen=True)
+class VolumeParseResult:
+    volumes: List[Volume]
+    secret_files: List[SecretFile]
+    disk_volumes: List[DiskVolume]
+
+
+@dataclass(frozen=True)
+class EnvParseResult:
+    env: Dict[str, str]
```
asvetlov, line 42: `VolumeParseResult` use the plural form for fields: `volumes`, `secret_files` etc.
What is the reason for the single form (`env`) in `EnvParseResult`?

romasku: By `env` I mean environment,  single form. I used plurals because `VolumeParseResult` contains `List` objects, but here is a single `Dict`. That was my point, but I can change to plurals if you think that way it will look cleaner.

asvetlov: Got you. Makes sense

asvetlov (approved):
LGTM

PR #1733, opened 2020-09-11
Use SI suffixes instead of GNU suffixes for memory sizes

romasku:
Switches from GNU suffixes to SI suffixes. More information [here](https://en.wikipedia.org/wiki/Binary_prefix). In short:
Suffixes like **M** or **G** should be used for decimal quantifiers and **Mi** or **Gi** (pronounced Mebi and Gibi) should be used for binary quantifiers.

### DIFF
--- a/neuromation/cli/parse_utils.py
+++ b/neuromation/cli/parse_utils.py
@@ -9,46 +9,44 @@ _T = TypeVar("_T")
 
 
 def parse_memory(memory: str) -> int:
-    """Parse string expression i.e. 16M, 16MB, etc
-    M = 1024 * 1024, MB = 1000 * 1000
+    """Parse memory string in SI format:
+    M = 10 ** 6, Mi = 2 ** 20
 
-    returns value in bytes"""
+    returns value in bytes
+    """
 
-    # Mega, Giga, Tera, etc
-    prefixes = "MGTPEZY"
-    value_error = ValueError(f"Unable parse value: {memory}")
-
-    if not memory:
-        raise value_error
-
-    pattern = r"^(?P<value>\d+)(?P<units>(kB|kb|K|k)|((?P<prefix>[{prefixes}])(?P<unit>[bB]?)))$".format(  # NOQA
-        prefixes=prefixes
-    )
-    regex = re.compile(pattern)
-    match = regex.fullmatch(memory)
-
-    if not match:
-        raise value_error
-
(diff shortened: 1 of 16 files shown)


tests/cli/formatters/test_storage_formatters.py
```diff
@@ -415,11 +415,11 @@ def test_long_formatter(self) -> None:
 
         formatter = LongFilesFormatter(human_readable=True, color=False)
         assert list(formatter(self.files_and_folders)) == [
-            "-r    2.0K 2018-01-01 03:00:00 File1",
-            "-r    1.0K 2018-10-10 13:10:10 File2",
-            "-r 1000.0K 2019-02-02 05:02:02 File3 with space",
-            "dm       0 2017-03-03 06:03:03 Folder1",
-            "dm       0 2017-03-03 06:03:02 1Folder with space",
+            "-r    2.0 Ki 2018-01-01 03:00:00 File1",
```
serhiy-storchaka, line 418: Why does it show "Ki" and not "KiB" as for disks formatter?

serhiy-storchaka (commented):
I am not sure that we should do it right now.

Advantages:
* More compatible with Kubernetes.
* Conforming to standards.

Disadvantages:
* Longer output (`"5 GiB"` vs `"5G"`, `"123 Bytes"` vs `"123B"`).
* Less compatible with GNU tools.

tests/cli/formatters/test_storage_formatters.py
romasku: It was striped in formatter (here is how it looks like on master brunch):
https://github.com/neuromation/platform-client-python/blob/18ecc3cf577a16c44df41fd24f14335e7cf10dc0/neuromation/cli/formatters/storage.py#L477-L478
Now I fixed that.

neuromation/cli/parse_utils.py
```diff
@@ -9,46 +9,44 @@
...
+    }
+    result: Optional[int] = None
+    try:
+        result = int(memory)
+    except ValueError:
+        for suffixes, factor in suffix_to_factor.items():
+            if result:
+                break
+            for suffix in suffixes:
+                if memory.endswith(suffix):
```
asvetlov, line 41: From the practical point of view, I suggest supporting case-insensitive comparison. `MiB`, `MIB`, and `mib` are equal, isn't it?

Issue #1735, opened 2020-09-11
Output after removing a disk


daria-moskovskykh:
There is no output after removing a disk.
See the screenshot https://monosnap.com/file/ggphK5dEMs41dypqDMXbSkLqZCdk8I

asvetlov:
@romasku please apply this tiny change

romasku:
I also checked how we handle other `rm` commands:
- `neuro secrets rm` has no output
- `neuro storage rm` only prints deleted files if `-v` (verbosity flag) is set
- And after https://github.com/neuromation/platform-client-python/pull/1738 `neuro disk rm` will always print "disk was removed".

Probably we should unify these behaviors. @asvetlov what do you think?

asvetlov:
I like printing when `-v` is set only everywhere.
Or, at least, we should skip printing if `-q` is passed.

PR #1738, opened 2020-09-14
Add disk was removed output

romasku:
Closes https://github.com/neuromation/platform-client-python/issues/1735

### DIFF
--- a/neuromation/cli/disks.py
+++ b/neuromation/cli/disks.py
@@ -82,7 +82,7 @@ async def rm(root: Root, disk_id: str) -> None:
     """
     await root.client.disks.rm(disk_id)
     if root.verbosity > 0:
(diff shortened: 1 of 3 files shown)


neuromation/cli/disks.py
```diff
@@ -78,8 +80,9 @@ async def rm(root: Root, disk_id: str) -> None:
     """
     Remove disk DISK_ID.
     """
-
     await root.client.disks.rm(disk_id)
+    if root.verbosity > 0:
+        click.echo(f"Disk with id: '{disk_id}' was successfully removed.")
```
serhiy-storchaka, line 85: I think `:` is not needed here grammatically.

> Disk with id 'abcd-1234' was successfully removed.

serhiy-storchaka (approved):
:+1:

PR #1743, opened 2020-09-15
Make "neuro image tags" outputting a list of tags instead of images

### LINKED ISSUE #1741: Make images.tags() returning a list of tags
After #899 `images.tags()` returns a list of images with tags instead of a list of tags (`List[RemoteImage]` instead of `List[str]`). It affected also the output of command `neuro image tags` which now outputs a list of full image URIs (unlike to `neuro image ls` which shorten URIs).

The documentation of command `neuro image tags` says only about a list of tags, not images. `images.tags()` is still not documented.

Returning a list of images from a method which does not contain "image" in name looks surprising to me. I think it would be better to make it returning a list of tags again. To help creating `RemoteImage` with tag we can add method `RemoteImage.with_tag()`.

serhiy-storchaka:
Closes  #1741.

See also alternate PR #1744.

### DIFF
--- a/CHANGELOG.D/1741.feature
+++ b/CHANGELOG.D/1741.feature
@@ -1,0 +1,1 @@
+Command `neuro image tags` outputs now a list of tags instead of a list of images.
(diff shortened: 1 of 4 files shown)


asvetlov:
I slightly prefer this PR over #1744 but have no strong opinion here

PR #1744, opened 2020-09-15
Make images.ls() returning a list of tags instead of images

### LINKED ISSUE #1741: Make images.tags() returning a list of tags

serhiy-storchaka:
Closes  #1741.

See also alternate PR #1743.

### DIFF
--- a/docs/images_reference.rst
+++ b/docs/images_reference.rst
@@ -66,6 +66,24 @@ Images
                *local* is ``None`` (:class:`LocalImage`)
 
 
+   .. comethod:: ls() -> List[RemoteImage]
+
+      List images on Neuro registry available to the user.
+
+      :return: list of remote images not including tags
+               (:class:`List[RemoteImage]`)
+
+
+   .. comethod:: tags(image: RemoteImage) -> List[RemoteImage]
+
+      List tags of the remote *image*.
+
+      :param RemoteImage image: a spec for remote image without tag on Neuro
+                                registry.
+
+      :return: list of tags (:class:`List[str]`)
+
+
 AbstractDockerImageProgress
 ===========================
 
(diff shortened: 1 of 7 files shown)


docs/images_reference.rst
```diff
@@ -66,6 +66,24 @@ Images
...
 
+   .. comethod:: ls() -> List[RemoteImage]
+
+      List images on Neuro registry available to the user.
+
+      :return: list of remote images not including tags
+               (:class:`List[RemoteImage]`)
+
+
+   .. comethod:: tags(image: RemoteImage) -> List[RemoteImage]
```
asvetlov, line 77:
```suggestion
   .. comethod:: tags(image: RemoteImage) -> List[str]
```

neuromation/api/images.py
```diff
@@ -172,19 +171,18 @@ def _validate_image_for_tags(self, image: RemoteImage) -> None:
         if not image.name:
             raise ValueError(err + "missing image name")
 
-    async def tags(self, image: RemoteImage) -> List[RemoteImage]:
+    async def tags(self, image: RemoteImage) -> List[str]:
```
asvetlov, line 174: If you are touching the method -- please convert it into async generator that yields strings (if you prefer it over RemoteImage, I have no strong opinion here).

Please apply the same transformation to `images.ls()`

PR #1745, opened 2020-09-17
Check URIs for allowed parts and support special characters in paths

### LINKED ISSUE #1531: neuro storage operations implicitly ignore argument part after `?` and `#`

serhiy-storchaka:
* Fragment, query, user, password and port parts are not allowed now
  in URIs. Only allowed parts are scheme, host and path.
* Supported characters like "%", "#", "?", "@", ":", "~" in local file paths.

Closes #1531

### DIFF
--- a/neuromation/api/parser.py
+++ b/neuromation/api/parser.py
@@ -106,7 +106,7 @@ class Parser(metaclass=NoPublicConstructor):
             raw_uri, container_path, read_only = self._parse_generic_volume(
                 volume, allow_rw_spec=True, resource_name="disk volume"
             )
-            disk_uri = self._parse_disk_resource(str(raw_uri))
+            disk_uri = self._parse_disk_resource(raw_uri)
             disk_volumes.append(DiskVolume(disk_uri, container_path, read_only))
         return disk_volumes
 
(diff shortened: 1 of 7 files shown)


neuromation/api/parser.py
```diff
@@ -85,7 +88,7 @@ def _build_secret_files(self, input_volumes: Set[str]) -> List[SecretFile]:
             raw_uri, container_path, _ = self._parse_generic_volume(
                 volume, allow_rw_spec=False, resource_name="secret file"
             )
-            secret_uri = self._parse_secret_resource(str(raw_uri))
+            secret_uri = self._parse_secret_resource(raw_uri)
```
romasku, line 91: There is almost same code in line 109, please remove `str` call there.

PR #1746, opened 2020-09-17
Make disk removal in e2e more reliable

romasku:


tests/e2e/test_e2e_jobs.py
```diff
@@ -1131,41 +1140,33 @@ def test_job_working_dir(helper: Helper) -> None:
...
-    )
+def test_job_disk_volume(
+    helper: Helper, disk_factory: Callable[[str], ContextManager[str]]
+) -> None:
 
-    out = captured.out
-    match = re.match("Job ID: (.+)", out)
-    assert match is not None, captured
-    job_id = match.group(1)
+    with disk_factory("1G") as disk:
```
serhiy-storchaka, line 1148: Can disk be deleted while it is still mounted in a job?

asvetlov: No, I believe.
Let's merge and monitor Github Checks

asvetlov (approved):
LGTM

Issue #1742, opened 2020-09-15
"neuro disk get" always outputs full URI


serhiy-storchaka:
There is a difference between output of `neuro disk ls` and `neuro disk get`.
```
$ neuro disk ls
Id                                         Storage  Uri                                                             Status
disk-05e70bdc-8db3-42cd-b1bd-d709c0bf5d30  2.0G     disk:/neuro-cli-e2e/disk-05e70bdc-8db3-42cd-b1bd-d709c0bf5d30   Ready 
...

$ neuro disk ls --full-uri
Id                                         Storage  Uri                                                                      Status
disk-05e70bdc-8db3-42cd-b1bd-d709c0bf5d30  2.0G     disk://default/neuro-cli-e2e/disk-05e70bdc-8db3-42cd-b1bd-d709c0bf5d30   Ready 
...

$ neuro disk get disk-05e70bdc-8db3-42cd-b1bd-d709c0bf5d30
Id                                         Storage  Uri                                                                     Status  Created at  Last used
disk-05e70bdc-8db3-42cd-b1bd-d709c0bf5d30  2.0G     disk://default/neuro-cli-e2e/disk-05e70bdc-8db3-42cd-b1bd-d709c0bf5d30  Ready   yesterday            
```
The former outputs shorten URI if possible and has option `--full-uri`. The latter always outputs full URI.

I think it would be better to shorten URI in the `neuro disk get` output and add support of option `--full-uri`.

PR #1747, opened 2020-09-17
Add support of --full-uri to neuro disk get

romasku:
Closes https://github.com/neuromation/platform-client-python/issues/1742


serhiy-storchaka (approved):
:+1:

PR #1753, opened 2020-09-23
Restore backward compatibility by adding back client.parse.env() method

asvetlov:


neuromation/api/parser.py
```diff
@@ -144,6 +145,16 @@ def _remote_to_local_image(self, image: RemoteImage) -> LocalImage:
         )
         return parser.convert_to_local_image(image)
 
+    def env(self, env: Sequence[str]) -> Tuple[Dict[str, str], Dict[str, URL]]:
```
serhiy-storchaka, line 148: Should not it have the `env_file` parameter?

serhiy-storchaka (approved):

serhiy-storchaka:
`neuro-flow` alos expects that `client.parse.volumes()` returns a 2-tuple. Maybe add `__iter__` to `VolumeParseResult` which returns `iter((self.volumes, self.secret_files))`?

neuromation/api/parser.py
asvetlov: Good catch. Added

asvetlov:
> neuro-flow alos expects that client.parse.volumes() returns a 2-tuple. Maybe add __iter__ to VolumeParseResult which returns iter((self.volumes, self.secret_files))

Thanks!
I've implemented `__len__()` / `__getitem__()` pair.

serhiy-storchaka (approved):

PR #1758, opened 2020-09-27
Added Keycloak support

dalazx:
KeyCloak strictly requires requests to be `application/x-www-form-urlencoded` encoded. Auth0 supports (and even suggests) according to their docs this encoding as well: https://auth0.com/docs/api/authentication#get-token


dalazx:
@dependabot rebase

asvetlov (approved):
lgtm

PR #1765, opened 2020-10-06
handle client connection error in neuro config show command

### LINKED ISSUE #1763: Show config even if there is no connection to cluster available
I guees this drawback was added after implementing the feature about showing available cluster resources.
We can display something like N/A in the table on connection error.

zubenkoivan:
Fixes #1763


asvetlov (approved):
LGTM.
Please add a changelog before merging

PR #1790, opened 2020-10-21
Fix bug that caused to kill jobs from other tests

romasku:
e2e/conftest.py uses some simple heuristic to catch `neuro run` like commands to kill jobs started by tests. Previously it joined all arguments by space and checked for some substrings. It did not working properly, because `'run'` is substring of `'jobs ls --status running'`.

Now it checks arguments directly, so it should work properly.

### DIFF
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -483,10 +483,7 @@ class Helper:
             raise
         out = proc.stdout
         err = proc.stderr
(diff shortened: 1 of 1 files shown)


asvetlov (approved):
Heh! Tricky!

asvetlov:
@romasku you are the Hero of the Day!

PR #1794, opened 2020-10-26
Bump aiohttp to 3.7.1

asvetlov:


neuromation/api/storage.py
```diff
@@ -121,7 +121,10 @@ def _uri_to_path(self, uri: URL, *, normalized: bool = False) -> str:
 
     def _set_time_diff(self, request_time: float, resp: aiohttp.ClientResponse) -> None:
         response_time = time.time()
-        server_timetuple = parsedate(resp.headers.get("Date"))
```
serhiy-storchaka, line 124: See also https://github.com/python/typeshed/issues/4713.

Although the behavior of `parsedate(None)` is not explicitly documented and therefore your code is more future proof.

tests/cli/test_main.py
```diff
@@ -21,7 +21,7 @@ def test_steal_config(
     monkeypatch: Any,
 ) -> None:
     patched_jobs_counts.return_value = {}
-    folder = shutil.move(nmrc_path, tmp_path / "orig-cfg")  # type: ignore
+    folder = shutil.move(str(nmrc_path), str(tmp_path / "orig-cfg"))
```
serhiy-storchaka, line 24: Why do you need to convert paths to str? Does not `shutil.move()` support the path protocol?

neuromation/api/storage.py
asvetlov: this was my thought

tests/cli/test_main.py
asvetlov: mypy had blamed that `move` requires `str`.
Perhaps it is a typeshed spike.
I did not investigate but quickly converted to the required type.
It is the test only, I wanted to make it pass quickly.

hauntsaninja: I think typeshed's annotations are working around https://bugs.python.org/issue32689 (fixed in 3.9, reflected in typeshed).

PR #1814, opened 2020-11-04
Use server side pass config

romasku:


neuromation/api/config_factory.py
```diff
@@ -141,6 +146,25 @@ async def login_with_token(
         )
         self._save(config)
 
+    async def login_with_passed_config(
+        self,
+        config_data: str,
```
asvetlov, line 151: Don't require `config_data` but get it from `os.environ` if not passed.

neuromation/api/config_factory.py
```diff
@@ -141,6 +146,25 @@ async def login_with_token(
...
+        *,
+        timeout: aiohttp.ClientTimeout = DEFAULT_TIMEOUT,
+    ) -> None:
+        try:
+            data = json.loads(config_data)
```
asvetlov, line 156: Please work with base64-encoded json string, not bare json

neuromation/cli/job.py
```diff
@@ -1190,6 +1191,10 @@ async def run_job(
     disk_volumes = volume_parse_result.disk_volumes
 
     if pass_config:
+        env_name = NEURO_PASSED_CONFIG
+        if env_name in env_dict:
+            raise ValueError(f"{env_name} is already set to {env_dict[env_name]}")
+        # Logic below is deprecated, only used to support images with old client
```
asvetlov, line 1197: Drop old logic entirely.
New client worth new image regeneration.

neuromation/cli/utils.py
```diff
@@ -612,7 +615,13 @@ def pager_maybe(
 
 
 def steal_config_maybe(dst_path: pathlib.Path) -> None:
```
asvetlov, line 617: Drop this function to fail fast if an outdated dependency tries to import the function.

Hmm, looks like `Factory.get()` should steal config by default to make the functionality usable.
Otherwise, all `api.get()` callers should care about handling `pass_config` and `login_with_passed_config()` explicitly.

asvetlov (commented):
I have a few  comments below

neuromation/cli/utils.py
romasku: OK, got it.
Then, do we need `login_with_passed_config` at all? If `Factory.get()` steels config by default, is see no reason for someone to call this function.
Also by implementing it this way (only inside `.get`), we can hide from user that we use ENV variable -- he/she still will be aware of it because of validation, but it will be impossible to write code that depends on this fact.

neuromation/api/config_factory.py
romasku: Done

neuromation/api/config_factory.py
romasku: Done

neuromation/cli/job.py
romasku: Done

neuromation/api/config_factory.py
serhiy-storchaka: Why?

docs/functions_reference.rst
```diff
@@ -186,6 +192,21 @@ Config Factory
        :param aiohttp.ClientTimeout timeout: optional timeout for HTTP operations, see
                                              also :ref:`timeouts`.
 
+   .. comethod:: login_with_passed_config( \
+                      config_data: str, \
```
asvetlov, line 196: Please sync the signature

asvetlov (approved):
LGTM except for the tiny Docs fix

asvetlov (approved):

asvetlov:
Everything is green. Awesome!

neuromation/api/config_factory.py
```diff
@@ -141,6 +150,32 @@ async def login_with_token(
...
+            token = data["token"]
+            cluster = data["cluster"]
+            url = URL(data["url"])
+        except (ValueError, KeyError):
+            raise ConfigError(f"Data in passed config is malformed: {config_data}")
+        await self.login_with_token(token, url=url, timeout=timeout)
+        client = await self.get(timeout=timeout)
+
+        await client.config.switch_cluster(cluster)
+        await client.close()
```
asvetlov, line 177: Even if the new code doesn't use the storage saved configuration for the login -- cleaning up the storage is still a good idea unless we stop pushing configuration files on the storage.
Please lookup if the config is present and clean it up -- as `steal_config_maybe()` did for the previous version.

asvetlov (approved):

PR #1818, opened 2020-11-06
Fix image rm test

anayden:


asvetlov (approved):
LGTM when green

tests/e2e/test_e2e_images.py
```diff
@@ -200,19 +200,24 @@ def test_images_delete(
     tag: str,
 ) -> None:
     image_no_tag = image.replace(f":{tag}", "")
```
romasku, line 202: I think problem can be because image is not uploaded to registry. `image` fixture only creates local image.

tests/e2e/test_e2e_images.py
```diff
@@ -200,19 +200,24 @@ def test_images_delete(
...
     matching_lines = [
-        line for line in captured.out.splitlines() if image_full_str == line.split()[0]
+        line
+        for line in captured.out.splitlines()
+        if image_full_str_no_tag == line.split()[0]
     ]
     assert len(matching_lines) == 1
 
-    helper.run_cli(["image", "rm", image])
+    helper.run_cli(["image", "rm", image_full_str])
```
romasku, line 220: `image_full_str` doesn't contain tag. 
```suggestion
    helper.run_cli(["image", "rm", f"{image_full_str}:{tag}"])
```

tests/e2e/test_e2e_images.py
```diff
@@ -200,19 +200,24 @@ def test_images_delete(
...
 
     captured = helper.run_cli(["image", "ls", "-l", "--full-uri"])
     matching_lines = [
```
romasku, line 223: I'm not sure that this check is good enough. You are only removing single tag, image repo (`image://default/neuro-e2e-user/e2e-banana-image` is content of `image_full_str`) can be still in the output of `neuro image ls` if tags generated by other tests are still there.

anayden: Right, image deletion works for single tag, that's expected

tests/e2e/test_e2e_images.py
anayden: Thanks! I somehow assumed `image` does the push, but never checked it.

tests/e2e/test_e2e_images.py
romasku: Sorry, it's wrong. Just forget.

PR #1826, opened 2020-11-09
Use better error reporting for exceptions with no args

romasku:


asvetlov (approved):

asvetlov:
Note, ETIMEDOUT is suppressed by d7427e2

PR #1827, opened 2020-11-10
Total running jobs quota

romasku:


neuromation/api/admin.py
```diff
@@ -131,8 +132,9 @@ async def set_user_quota(
         self,
         cluster_name: str,
         user_name: str,
-        gpu_value_minutes: Optional[float],
-        non_gpu_value_minutes: Optional[float],
```
romasku, line 135: Quotas are ints everywhere (on server, in cli parser), so I think `Optional[float]` was a mistake.

neuromation/cli/job.py
```diff
@@ -270,6 +270,12 @@ def job() -> None:
     show_default=True,
     help="Upload neuro config to the job",
 )
+@option(
+    "--wait-for-job-quota/--no-wait-for-job-quota",
```
asvetlov, line 274: The name is quite long.
I have no idea what should be better.
Maybe @mariyadavydova can suggest something?

romasku: I can suggest the following alternatives:

`--wait-for-quota`
`--queue-job`
`--fail-on-quota/--no-fail-on-quota`
`--keep-pending`
`--no-quota-error`
`--wait-for-seat`
`--run-eventually`

But as for me, none of them are good.

mariyadavydova: I like `--wait-for-seat`.

asvetlov: Ok, let's rename to `--wait-for-seat` and merge

asvetlov (approved):

PR #1831, opened 2020-11-12
Use partial download/upload when resume copying

serhiy-storchaka:


neuromation/api/utils.py
```diff
@@ -29,6 +28,8 @@
 else:
     from async_generator import asynccontextmanager
 
+asynccontextmanager
```
asvetlov, line 31: Is the line required?

asvetlov (approved):
LGTM except a tiny question

neuromation/api/utils.py
serhiy-storchaka: Yes, it silences warning "imported but not used". `asynccontextmanager` is imported from `utils` in several other places, so I just add a workaround. In different PR I'll propose better solution.

Will add a comment.

asvetlov (approved):

PR #1841, opened 2020-11-17
Add option --continue for "neuro cp".

serhiy-storchaka:
It specified, copy only the part of the source file
past the end of the destination file and append it
to the destination file if the destination file is
newer and not longer than the source file.
Otherwise copy the whole file.


asvetlov (approved):
LGTM

PR #1842, opened 2020-11-17
Add data generator for storage API performance tests.

YevheniiSemendiak:
Save small script for generating specified number of files with specified aggregated size for future uses.

@serhiy-storchaka, it might be handy for you, while you're fixing the storageAPI performance issues.
Also, I could change the way a filename is selected. Please, let me know if it is needed.

cc @mariyadavydova

### DIFF
--- a/build-tools/garbage-files-generator.py
+++ b/build-tools/garbage-files-generator.py
@@ -8,19 +8,13 @@ import pathlib
 import re
 from functools import reduce
 
+from rich.progress import Progress
+
 
 logging.basicConfig(
     level=logging.INFO,
     format="%(asctime)s - %(levelname)-8s - %(filename)s:%(lineno)d |> %(message)s",
 )
-
-try:
-    import tqdm
-except ImportError:
-    tqdm = None
-    logging.warning(
-        f"tqdm library is not available, progress bar will not be displayed."
-    )
 
 
 OUTPUT_FOLDER = pathlib.Path("./data")
@@ -63,41 +57,40 @@ def generate_data(
             f" The resulting dataset will be {resulting_size} bytes"
             f" ({round(resulting_size / total_size, 2)} times larger)."
         )
-    if tqdm is not None:
-        progress_bar = tqdm.tqdm(total=files_count)
 
     created_files = 0
     folders_counter = 0
-    while created_files < files_count:
-        files_count_z = str(folders_counter).zfill(name_length * tree_depth)
+    with Progress() as progress:
+        data_gen_task = progress.add_task(
+            "[green]Generating data...", total=files_count
+        )
+        while created_files < files_count:
+            files_count_z = str(folders_counter).zfill(name_length * tree_depth)
 
-        split_path = []
-        for level in range(tree_depth):
-            split_path.append(
-                files_count_z[name_length * level : name_length * (level + 1)]
+            split_path = []
+            for level in range(tree_depth):
+                split_path.append(
+                    files_count_z[name_length * level : name_length * (level + 1)]
+                )
+
+            folder_path = reduce(
+                lambda parent, child: parent / child, (OUTPUT_FOLDER, *split_path)
             )
 
-        folder_path = reduce(
-            lambda parent, child: parent / child, (OUTPUT_FOLDER, *split_path)
-        )
+            folder_path.mkdir(parents=True, exist_ok=True)
+            folders_counter += 1
 
-        folder_path.mkdir(parents=True, exist_ok=True)
-        folders_counter += 1
-
-        for i in range(branching_factor):
-            if created_files < files_count:
-                file_name = str(i).zfill(name_length)
-                full_file_name = folder_path / file_name
-                with full_file_name.open("wb") as file:
-                    file.write(os.urandom(file_size_bytes))
-                created_files += 1
-                if tqdm is not None:
-                    progress_bar.update(1)
-            else:
-                break
-    if tqdm is not None:
-        progress_bar.close()
-    logging.info("Data generation completed.")
+            for i in range(branching_factor):
+                if created_files < files_count:
+                    file_name = str(i).zfill(name_length)
+                    full_file_name = folder_path / file_name
+                    with full_file_name.open("wb") as file:
+                        file.write(os.urandom(file_size_bytes))
+                    created_files += 1
+                    progress.advance(data_gen_task)
+                else:
+                    break
+        logging.info("Data generation completed.")
 
 
 def _parse_args() -> argparse.Namespace:
--- a/build-tools/garbage-files-generator.py
+++ b/build-tools/garbage-files-generator.py
@@ -1,22 +1,13 @@
 #!/usr/bin/env python
 
 import argparse
-import logging
 import math
 import os
 import pathlib
 import re
 
+from rich.console import Console
 from rich.progress import Progress
-
-
-logging.basicConfig(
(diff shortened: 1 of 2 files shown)


asvetlov (changes requested):

build-tools/garbage-files-generator.py
```diff
@@ -0,0 +1,123 @@
...
+from functools import reduce
+
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s - %(levelname)-8s - %(filename)s:%(lineno)d |> %(message)s",
+)
+
+try:
+    import tqdm
```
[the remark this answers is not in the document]
YevheniiSemendiak: That's a very good suggestion, thank you!
Replaced in  410632d

build-tools/garbage-files-generator.py
```diff
@@ -0,0 +1,123 @@
...
+    name_length = len(str(branching_factor))
+    logging.info(
+        f"Generating {files_count} files {file_size_bytes} bytes each, "
+        f"which in total: {total_size} bytes."
+    )
+
+    OUTPUT_FOLDER.mkdir(parents=True, exist_ok=True)
+    fs_stats = os.statvfs(OUTPUT_FOLDER)
+    if file_size_bytes < 10 * fs_stats.f_frsize:
+        resulting_size = (
```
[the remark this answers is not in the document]
YevheniiSemendiak: Ok, added just in case if someone will be curious why the generated dataset size is larger than requested. But in general, you're right, it will not impact the generation performance, removed in 4955b95

build-tools/garbage-files-generator.py
```diff
@@ -0,0 +1,123 @@
...
+
+        folder_path.mkdir(parents=True, exist_ok=True)
+        folders_counter += 1
+
+        for i in range(branching_factor):
+            if created_files < files_count:
+                file_name = str(i).zfill(name_length)
+                full_file_name = folder_path / file_name
+                with full_file_name.open("wb") as file:
+                    file.write(os.urandom(file_size_bytes))
```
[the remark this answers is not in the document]
YevheniiSemendiak: that's a good catch, thank you for it!
replaced with a buffered bytearray 🙂  in b52e280

build-tools/garbage-files-generator.py
```diff
@@ -0,0 +1,118 @@
...
+
+            split_path = []
+            for level in range(tree_depth):
+                split_path.append(
+                    files_count_z[name_length * level : name_length * (level + 1)]
+                )
+
+            folder_path = reduce(
+                lambda parent, child: parent / child, (OUTPUT_FOLDER, *split_path)
+            )
```
serhiy-storchaka, line 72:
```suggestion
            folder_path = OUTPUT_FOLDER.joinpath(*split_path)
```

build-tools/garbage-files-generator.py
```diff
@@ -0,0 +1,118 @@
...
+    file_size_bytes = math.ceil(total_size / files_count)
+    tree_depth = math.floor(math.log(files_count, branching_factor))
+    name_length = len(str(branching_factor))
+    logging.info(
+        f"Generating {files_count} files {file_size_bytes} bytes each, "
+        f"which in total: {total_size} bytes."
+    )
+
+    buffer_size = min(file_size_bytes, 16 * 2 ** 20)  # 16MB at max
+    garbage = bytearray(os.urandom(buffer_size))
```
serhiy-storchaka, line 46: `bytearray()` is redundant.

build-tools/garbage-files-generator.py
```diff
@@ -0,0 +1,118 @@
...
+    parser.add_argument("total_size", type=_parse_size)
+    parser.add_argument("--branching-factor", type=int, default=100)
+    return parser.parse_args()
+
+
+def _parse_size(size: str) -> int:
+    units = {"B": 1, "KB": 2 ** 10, "MB": 2 ** 20, "GB": 2 ** 30, "TB": 2 ** 40}
+    size = size.upper()
+    # print("parsing size ", size)
+    if not re.match(r" ", size):
```
serhiy-storchaka, line 111: It just checks here that `size` starts with `" "`. You perhaps mean `not re.search(r" ", size)` or better `" " not in size`.

But it is better to use regular expression for parsing:

```python
number, unit = re.fullmatch(r"(.*)([KMGT]?B)", size.strip()).groups(1, 2)
```

serhiy-storchaka:
My internet connection is slow. @YevheniiSemendiak could you please test performance with new client? It should better resume copying now, and in case it still fails, there is a new option `--continue` which allows to continue with other run.

build-tools/garbage-files-generator.py
YevheniiSemendiak: thanks! removed in c2a775c

build-tools/garbage-files-generator.py
YevheniiSemendiak: applied re as proposed in c2a775c

YevheniiSemendiak:
@serhiy-storchaka, OK, I will try how it works within the cluster (from one job with --pass-config -> storage), since my channel is not that wide for upload too 🙂

build-tools/garbage-files-generator.py
```diff
@@ -0,0 +1,113 @@
...
+    parser.add_argument("total_size", type=_parse_size)
+    parser.add_argument("--branching-factor", type=int, default=100)
+    return parser.parse_args()
+
+
+def _parse_size(size: str) -> int:
+    units = {"B": 1, "KB": 2 ** 10, "MB": 2 ** 20, "GB": 2 ** 30, "TB": 2 ** 40}
+    size = size.upper()
+    # print("parsing size ", size)
+    number, unit = re.fullmatch(r"(\d*)([KMGT]?B)", size.strip()).groups()
```
serhiy-storchaka, line 108:
```suggestion
    number, unit = re.fullmatch(r"(\d+(?:\.\d*))([KMGT]?B)", size.strip()).groups()
```
If you want to support `1.2GB`.

serhiy-storchaka (approved):

build-tools/garbage-files-generator.py
YevheniiSemendiak: Good idea, thanks again for your help!

YevheniiSemendiak:
@asvetlov, I've taken over the benchmarks of the storage API. Should we keep this PR open, or let's merge it?

asvetlov:
Up to you. I'm fine with merging since Serhiy approved the PR

build-tools/garbage-files-generator.py
```diff
@@ -0,0 +1,113 @@
...
+                                completed=iteration,
+                                description=f"[cyan]Writing file {full_file_name}...",
+                            )
+                        if tail_size != 0:
+                            file.write(garbage[:tail_size])
+                    created_files += 1
+                    progress.advance(data_gen_task)
+                else:
+                    break
+    logging.info("Data generation completed.")
```
asvetlov, line 92: Very minor: if you started to use rich -- please create a `rich.console.Console` and use `console.log()` instead of `logging.info()`.  Mixing two styles can make a weird result eventually.

asvetlov (approved):

build-tools/garbage-files-generator.py
YevheniiSemendiak: Good idea, I replaced logging with rich.Console in 8c089ba.

PR #1847, opened 2020-11-18
Support specifying active cluster in local .neuro.toml

### LINKED ISSUE #1846: Support per-project setting of active cluster
It should be possible to specify active cluster name in local configuration file `.neuro.toml`. If it is not set, the global setting should be used.

serhiy-storchaka:
Closes #1846.


asvetlov (approved):
LGTM, let's merge and release!

PR #1859, opened 2020-11-20
Display restarts in cli when there was restarts

romasku:


asvetlov (approved):
Ok.
Did you have a feature request or just make the implementation similar to other properties displaying?

P.S.
Add CHANGELOG please

romasku:
> Did you have a feature request or just make the implementation similar to other properties displaying?

There was a discussion in slack about preemptible jobs (and how to get a number of restarts for them). Then I updated the server to count include SUSPENDED as restart in https://github.com/neuro-inc/platform-api/pull/1399, and this is PR to the client that makes this functionality visible in CLI.

asvetlov:
ok, thanks

PR #1870, opened 2020-11-26
Add `-f` flag to `neuro image rm`

### LINKED ISSUE #1828: Delete all tags before deleting image with `neuro images rm`

anayden:
Closes #1828 

If there's more than 1 tag for image, the CLI checks whether any other tags reference the same image and if that's the case refuses to delete image without the `-f` flag.

### DIFF
--- a/CHANGELOG.D/1828.feature
+++ b/CHANGELOG.D/1828.feature
@@ -1,0 +1,1 @@
+Add `-f` flag to `neuro image rm` to force delete images that have multiple tag references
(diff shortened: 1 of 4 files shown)


asvetlov (changes requested):
Please document new API

anayden:
@asvetlov there's no new API, only changes in CLI behavior. There was an unused CLI method that I've deleted.

asvetlov (commented):
Looks good.
CHANGELOG is required though

asvetlov (approved):

PR #1880, opened 2020-12-03
Add disk life span info to CLI disk formatters.

YevheniiSemendiak:
Useful, if one wants to know when the disk will disappear.

Actually, it was a feature request from the Synthesis team.
cc @Rauf-Kurbanov


asvetlov:
Good idea

romasku (approved):
Looks good to me

PR #1895, opened 2020-12-09
Make str(image) roundtrip with the parser

serhiy-storchaka:
It now %-encode special and non-ascii characters in the image URI.


asvetlov (approved):

asvetlov:
Add changelog please

asvetlov:
@serhiy-storchaka please sync with the master.
Packages are renamed but I believe git will merge the upstream without significant problems.

PR #1900, opened 2020-12-11
Make isort for tests stable.

serhiy-storchaka:


asvetlov (approved):
LGTM!

I wanted to propose the same change.

PR #1978, opened 2021-02-08
support version check in custom pypi server

zubenkoivan:


neuro-cli/src/neuro_cli/version_utils.py
```diff
@@ -120,7 +123,12 @@ def _read_package(db: sqlite3.Connection, package: str) -> Optional[Record]:
 async def _fetch_package(
     session: aiohttp.ClientSession, package: str
 ) -> Optional[Record]:
-    url = URL(f"https://pypi.org/pypi/{package}/json")
```
asvetlov, line 123: I believe you can just patch this line when making a custom archive; there is no need in the source code modification

PR #1993, opened 2021-02-15
Print job id on detach

romasku:
Closes https://github.com/neuro-inc/platform-client-python/issues/1930

### DIFF
--- a/CHANGELOG.D/1993.feature
+++ b/CHANGELOG.D/1993.feature
@@ -1,0 +1,1 @@
+Added printing of job id when `neuro flow` or `neuro attach` exits.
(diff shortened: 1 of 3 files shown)


asvetlov (changes requested):
`JobStopProgress` should be modified in a similar way probably.

`DetailedJobStopProgress.tick()` has `msg = yes() + " Stopped"` line along with `DetailedExecStopProgress.tick()`

romasku:
Sorry, I can't find a place that I missed updating.
I update both for `DetailedJobStopProgress` and `DetailedExecStopProgress`. The first update is line 714 (the change is tiny because `JobDescription` is available in `tick()`) and all other changed are related to `DetailedExecStopProgress`.

asvetlov (approved):
Sorry, my bad.
LGTM after adding the changelog

PR #2004, opened 2021-02-19
Increase read buffer

romasku:


asvetlov (approved):
LGTM.
Please add a change note before the merging

PR #2034, opened 2021-03-10
Enable soft wrap for generate-run-command

YevheniiSemendiak:
If not enabled, Rich will add `\n`s implicitly and one is not able to `copy -> paste -> redact -> run` command in terminal.


romasku (approved):
Please add changelog (I think this should go to a `.bugfix`)

PR #2048, opened 2021-03-18
Add recovery after config issue

romasku:
The recovery is done on the "neuro-sdk" side because in other cases we would have to duplicate "try ... except ConfigError: recover()" code in CLI, neuro-flow, and neuro-extras.


asvetlov (approved):
LGTM, add changelog please

PR #2053, opened 2021-03-22
Use more precise datetime formatting in job status

romasku:
Closes https://github.com/neuro-inc/platform-client-python/issues/2043


Makefile
```diff
@@ -1,6 +1,6 @@
 SHELL := /bin/bash
 
-PYTEST_ARGS=
+PYTEST_ARGS= --rich-gen
```
asvetlov, line 3: Why the flag is always on? OMHO it can modify generated files during the test run on CI without letting us know.

asvetlov (changes requested):
Add changelog please

Makefile
romasku: I just forgot to remove it, sorry.

asvetlov (approved):
LGTM, thanks!

PR #2056, opened 2021-03-22
Do print about removed disk by default

romasku:
Closes https://github.com/neuro-inc/platform-client-python/issues/2024


asvetlov (approved):
Changelog, please

PR #2086, opened 2021-04-12
Fix attach to job that exits immediately

romasku:


neuro-cli/src/neuro_cli/ael.py
```diff
@@ -255,6 +256,15 @@ async def _process_attach_single_try(
...
                     progress.detach(job)
                     sys.exit(0)
+        except WSServerHandshakeError as e:
+            # Websocket handshake error has no access to response body, so we can only
+            # check the status here. Status 400 can mean:
+            # - wrong job id (cannot happen here as it is checked above)
+            # - wrong usage api usage (checked by tests)
+            # - container already stopped, so we can ignore such error
+
+            if e.status != 400:
```
romasku, line 266: We can also use special HTTP status for this case (`409 CONFLICT`or `410 GONE` may be relevant), but it is somewhat hacky.

asvetlov: `410 Gone` or sounds really good. I'd like to distinguish 'bad request data' and 'job is already finished'.

Maybe `404 Not Found` is the better alternative? At least we use this status for similar situations, isn't it?

romasku: Ok, `404` also sounds like a good idea to me. Sometimes the best option is the simplest one:)

romasku: PR to platform monitoring: https://github.com/neuro-inc/platform-monitoring/pull/377

asvetlov (approved):

PR #2090, opened 2021-04-13
Support formatting several values in one table cell

### LINKED ISSUE #2062: Support merging multiple cells in one column
Here is an example of the output of `neuro ps` (terminal width is 120):
```
  ID                                         NAME          STATUS    WHEN           IMAGE         OWNER   COMMAND       
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 
  job-9a90c85e-27a3-442a-9fb2-a493ae8377de   test-job-1…   running   10 seconds     ubuntu:lat…   YOU     bash -c       
                                                                     ago                                  'while true;  
                                                                                                          do sleep 0;   
                                                                                                          done'         
```
Most jobs do not have name, and the column NAME is just a waste of horizontal space for them. But if the job has name, the width of the NAME column can be not enough and the name will be truncated.

Fields containing spaces (like WHEN and COMMAND) can be wrapped and output as several lines. COMMAND is often long, so it is likely that it will take several lines even on wide screen. Therefore it is often the row takes several lines on terminal.

I propose to allow to merge some cells. For example, ID and NAME can be merged in one column and output one below other. Other good candidates are STATUS and WHEN: WHEN is the time of changing STATUS. It will save horizontal space. The resulting output can be:
```
  ID/NAME                                    STATUS/WHEN      IMAGE           OWNER   COMMAND
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
  job-9a90c85e-27a3-442a-9fb2-a493ae8377de   running          ubuntu:latest   YOU     bash -c 'while true; do sleep 0;
  test-job-12345                             10 seconds ago                           done'
```
BTW the same can be done on Web.

serhiy-storchaka:
Closes #2062.


asvetlov (approved):
LGTM

PR #2100, opened 2021-04-21
pass trace configs to client

zubenkoivan:


neuro-sdk/src/neuro_sdk/__init__.py
```diff
@@ -154,13 +154,20 @@
 
 
 def get(
-    *, path: Optional[Path] = None, timeout: aiohttp.ClientTimeout = DEFAULT_TIMEOUT
+    *,
+    path: Optional[Path] = None,
+    timeout: aiohttp.ClientTimeout = DEFAULT_TIMEOUT,
+    trace_configs: Optional[List[aiohttp.TraceConfig]] = None,
```
asvetlov, line 160: Where the argument is used/passed?

zubenkoivan: It will be used on server side. For example `platform-monitoring` used `neuro-sdk` to retrieve jobs

asvetlov: Ok

asvetlov (approved):

PR #2155, opened 2021-06-04
Fix loop close errors

romasku:
Current code sometimes triggers following tracebacks:
```
Exception ignored in: <function BaseEventLoop.__del__ at 0x101c69830>
Traceback (most recent call last):
  File "/usr/local/opt/python@3.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/base_events.py", line 628, in __del__
    self.close()
  File "/usr/local/opt/python@3.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/unix_events.py", line 54, in close
    super().close()
  File "/usr/local/opt/python@3.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/selector_events.py", line 91, in close
    self._close_self_pipe()
  File "/usr/local/opt/python@3.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/selector_events.py", line 98, in _close_self_pipe
    self._remove_reader(self._ssock.fileno())
  File "/usr/local/opt/python@3.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/selector_events.py", line 271, in _remove_reader
    key = self._selector.get_key(fd)
  File "/usr/local/opt/python@3.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/selectors.py", line 190, in get_key
    return mapping[fileobj]
  File "/usr/local/opt/python@3.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/selectors.py", line 71, in __getitem__
    fd = self._selector._fileobj_lookup(fileobj)
  File "/usr/local/opt/python@3.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/selectors.py", line 225, in _fileobj_lookup
    return _fileobj_to_fd(fileobj)
  File "/usr/local/opt/python@3.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/selectors.py", line 42, in _fileobj_to_fd
    raise ValueError("Invalid file descriptor: {}".format(fd))
ValueError: Invalid file descriptor: -1
```
This is because call to `asyncio.get_event_loop()` at line 63 creates one more event loop (if it is the main thread) and it isn't closed properly because we do not restore it.

### DIFF
--- a/neuro-cli/src/neuro_cli/asyncio_utils.py
+++ b/neuro-cli/src/neuro_cli/asyncio_utils.py
@@ -7,7 +7,7 @@ import ssl
 import sys
 import threading
 import warnings
-from asyncio.events import AbstractEventLoop
+from asyncio import AbstractEventLoop
 from concurrent.futures import ThreadPoolExecutor
 from types import TracebackType
 from typing import Any, Awaitable, Callable, Dict, List, Optional, Type, TypeVar
--- a/neuro-cli/src/neuro_cli/asyncio_utils.py
+++ b/neuro-cli/src/neuro_cli/asyncio_utils.py
@@ -7,6 +7,7 @@ import ssl
 import sys
 import threading
 import warnings
+from asyncio.events import AbstractEventLoop
 from concurrent.futures import ThreadPoolExecutor
 from types import TracebackType
 from typing import Any, Awaitable, Callable, Dict, List, Optional, Type, TypeVar
@@ -30,6 +31,7 @@ class Runner:
         self._started = False
         self._stopped = False
         self._executor = ThreadPoolExecutor()
+        self._old_loop: Optional[AbstractEventLoop] = None
         self._loop = asyncio.new_event_loop()
         self._loop.set_default_executor(self._executor)
         _setup_exception_handler(self._loop, self._debug)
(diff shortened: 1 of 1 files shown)


serhiy-storchaka (commented):
Could you please test it with Python 3.10? `get_event_loop()` will emit a warning when called outside of the event loop, and `set_event_loop()` will not always help.

neuro-cli/src/neuro_cli/asyncio_utils.py
```diff
@@ -7,6 +7,7 @@
 import sys
 import threading
 import warnings
+from asyncio.events import AbstractEventLoop
```
asvetlov, line 10:
```suggestion
from asyncio import AbstractEventLoop
```

asvetlov (approved):
I think the PR is good.
To avoid a deprecation warning the direct call to `get_event_loop_policy().get_event_loop()` can be used.

serhiy-storchaka (approved):

serhiy-storchaka:
What was the most common way to trigger that error?

romasku:
It often happened in neuro-flow when the following code is running:
https://github.com/neuro-inc/neuro-flow/blob/8d265db2ac43f1aafecaa7487e40197919a64f1d/neuro_flow/cli/click_types.py#L32-L33
(for example, the `neuro-flow restart` triggered it with an ~80% chance). When I patched code locally, the problem stopped reproducing.
I think that this example should trigger it:
```
async def noop():
    pass

with Runner() as runner:
    runner.run(noop)
```

I never saw this error in `neuro` cli directly.

PR #2190, opened 2021-06-29
Properly close asynchronous generators.

serhiy-storchaka:


serhiy-storchaka:
A workaround for Python bug related to GeneratorExit has been removed because it conflicts with `aclosing()`. The original bug had been fixed in upstream.

serhiy-storchaka:
In the following PR I am going to add a decorator which turns generators into context managers (the problem is with annotating) and document preferable use of generators.

asvetlov (approved):

PR #2201, opened 2021-07-06
Add a neuro storage du command

romasku:


asvetlov (approved):
LGTM but please rename the command to `df`; it works more similar to the standard `df` than `du`. E.g., you cannot calculate the size of particular folder; only the total NFS storage size.

asvetlov (approved):

PR #2211, opened 2021-07-14
Add file with full logs of each run

romasku:


neuro-cli/src/neuro_cli/log_formatter.py
```diff
@@ -1,11 +1,12 @@
+import copy
 import logging
 import sys
 
 from rich.console import Console
 
 
 class ConsoleHandler(logging.Handler):
-    def __init__(self, color: bool) -> None:
+    def __init__(self, color: bool, show_traceback: bool = False) -> None:
```
romasku, line 9: The default value is for `neuro-flow` as it also uses this class.

asvetlov (approved):

PR #2225, opened 2021-07-22
Cluster cli option

romasku:
This allows running some operations for a specific cluster instead of the current one. This can be helpful when the user wants to write a script that migrates a lot of data from one cluster to another.

### DIFF
--- a/CHANGELOG.D/2225.feature
+++ b/CHANGELOG.D/2225.feature
@@ -1,0 +1,2 @@
+Added `--cluster` option to `neuro run` command and `neuro disk`, `neuro secret` command groups. This option
+allows to perform actions for a specific cluster instead of current one.
--- a/CLI.md
+++ b/CLI.md
@@ -2120,6 +2120,7 @@ neuro secret ls [OPTIONS]
 Name | Description|
 |----|------------|
 |_--help_|Show this message and exit.|
+|_--cluster CLUSTER_|Look on a specified cluster \(the current cluster by default).|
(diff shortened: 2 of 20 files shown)


serhiy-storchaka (approved):

serhiy-storchaka:
I wonder if it would be better to save cluster name as a state of `Parser` and add a `Parser` metthod which creates a copy of `Parser` with different default cluster. Then instead
```python
volume_parse_result = root.client.parse.volumes(volume, cluster_name=cluster_name)
```
we would write
```python
parser = root.client.parse.with_cluster(cluster_name)
volume_parse_result = parser.volumes(volume)
```
or
```python
parser = root.client.parse.replace(cluster_name=cluster_name)
volume_parse_result = parser.volumes(volume)
```
What looks better to you?

romasku:
The solution with the direct argument (`.some_method(..., cluster_name=cluster_name)`) is the only option for some cases (jobs, disks, secrets), so I prefer to use also in parser to make API more consistent. Also, it's can be unexpected that `client.parser.with_cluster(client.cluster_name)` is not no-op - it "freezes" default cluster in returned parser object.

While implementing this PR, I thought about adding a similar method to the client itself, like:
```
with client.use_as_default_cluster(tmp_cluster_name):
    # Here current cluster is changed but only in memory (no config rewrite)
# default cluster is restored
```
But I rejected this idea as it is too tricky and too implicit.

serhiy-storchaka (approved):

asvetlov (approved):

PR #2229, opened 2021-07-26
Fix set_user_quota if credits == 0

YevheniiSemendiak:
In this case, we set credits to 'infinity', since [this](https://github.com/neuro-inc/platform-client-python/compare/ys/fix_set_usr_credits?expand=1#diff-88fbede459d7e6bdef38fc1ed9cdeb85e143080e793cadfe96ecb33a5e34155cR173) line evaluates to None, which is improper.


romasku (approved):
Thanks!

romasku:
Please add a `bugfix` changelog entry before merging.

PR #2237, opened 2021-08-03
Add possiblity to disable pypi version check via setting env var

YevheniiSemendiak:
We need to suppress version checks in different tools, which might have a lag before the release with the latest versions (such as neuromation base environment docker image, neuro-web-shell, etc.)

As an alternative - we could also utilize the [auto_envvar_prefix](https://click.palletsprojects.com/en/8.0.x/options/#dynamic-defaults-for-prompts) feature from click, but I'm not sure whether we need that thing at the moment.

### DIFF
--- a/CHANGELOG.D/2237.feature
+++ b/CHANGELOG.D/2237.feature
@@ -1,0 +1,1 @@
+Added possibility to disable PyPi version check via environment variable 'NEURO_CLI_DISABLE_PYPI_VERSION_CHECK'.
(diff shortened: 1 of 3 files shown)


asvetlov (changes requested):
Please use `envvar="NEURO_CLI_DISABLE_PYPI_VERSION_CHECK"` instead and `show_envvar=True`.

It makes the env var ustodiscoverable in generated docs

YevheniiSemendiak:
> Please use `envvar="NEURO_CLI_DISABLE_PYPI_VERSION_CHECK"` instead and `show_envvar=True`.
> 
> It makes the env var ustodiscoverable in generated docs

Pitty I didn't find such config by myself, thanks for the hint!
Done in c8498a5 
cc @asvetlov

asvetlov (approved):
LGTM.
Please write `CHANGELOG.D/2237.feature` file and click 'merge' when ready.

PR #2247, opened 2021-08-11
Blob commands

romasku:


asvetlov (approved):
Looks good.
Please postpone the merge until the service is deployed on staging.

Also, you might be interested in supporting filename autocompletion from #2248 before the merge.

PR #2251, opened 2021-08-13
Output debugging information for logs with the -v option.

serhiy-storchaka:


asvetlov (approved):
LGTM if applied for `-vv` mode only

PR #2254, opened 2021-08-17
handle job exit immediately after start

zubenkoivan:


serhiy-storchaka (commented):
Why not handle it inside `process_attach()`?

zubenkoivan:
> Why not handle it inside `process_attach()`?

Thanks. Actually found a better place for this.

serhiy-storchaka (approved):
:+1:

asvetlov (approved):

PR #2258, opened 2021-08-19
New blob storage

romasku:


asvetlov (commented):
Can we consider the merging of blob and bucket groups?
Flat is better than nested :)

romasku:
> ...the merging of blob and bucket groups
Thanks, it is good idea to keep all blob/buckets command in one place. We can do such renaming:
`neuro bucket create` -> `neuro blob create-bucket`
`neuro bucket ls` -> `neuro blob list-buckets`
`neuro bucket rm` -> `neuro blob rm-bucket`
`neuro bucket get` -> `neuro blob get-bucket`
Bucket-level operations should be rare, so it's probably OK to have longer names for them. If you had a different solution in mind, please share.

Also, I'm going to add some commands to the manager persistent credentials to buckets, they will also need (create/list/get/delete) commands, I think they can be named as `neuro blob create-buckets-credential` and so on.

PR #2269, opened 2021-08-25
Fix neuro project init tests

YevheniiSemendiak:


YevheniiSemendiak:
done in 67ee769

romasku:
Your solution is much cleaner than mine. Lets merge it instead.

romasku (approved):

PR #2273, opened 2021-08-27
Enable autocomplete for blob storage uris

romasku:
Code is a little bit too generic because I think we should enable autocomplete for all schemas when we will have time. This will help with resource sharing.

### DIFF
--- a/neuro-cli/src/neuro_cli/blob_storage.py
+++ b/neuro-cli/src/neuro_cli/blob_storage.py
@@ -22,7 +22,7 @@ from neuro_cli.click_types import (
     BUCKET_CREDENTIAL,
     BUCKET_NAME,
     CLUSTER,
-    StoragePathType,
+    PlatformURLType,
 )
 from neuro_cli.formatters.bucket_credentials import (
     BaseBucketCredentialsFormatter,
@@ -203,7 +203,7 @@ async def rmbucket(root: Root, cluster: Optional[str], buckets: Sequence[str]) -
 @click.argument(
     "paths",
     nargs=-1,
-    type=StoragePathType(allowed_schemes=["blob"]),
+    type=PlatformURLType(allowed_schemes=["blob"]),
 )
 @option(
     "--human-readable",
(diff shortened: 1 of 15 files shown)


neuro-sdk/src/neuro_sdk/buckets.py
```diff
@@ -618,13 +618,15 @@ async def _get_bucket_fs(
         async with self._get_provider(bucket_name, cluster_name) as provider:
             yield BucketFS(provider)
 
-    def _split_blob_uri(self, uri: URL) -> Tuple[str, str, str]:
+    def split_blob_uri(self, uri: URL) -> Tuple[str, str, str]:
```
asvetlov, line 621: Should the method be moved into `client.parse` namespace along with other URI helpers.

neuro-cli/src/neuro_cli/blob_storage.py
```diff
@@ -267,8 +273,13 @@ async def ls(
 
 @command()
 @option("--full-uri", is_flag=True, help="Output full bucket URI.")
-@click.argument("patterns", nargs=-1, required=False)
-async def glob(root: Root, full_uri: bool, patterns: Sequence[str]) -> None:
+@click.argument(
+    "patterns",
+    nargs=-1,
+    required=False,
+    type=PlatformURLType(allowed_schemes=["blob"]),
```
asvetlov, line 280: Very minor: we use URI instead of URL for platform things.

neuro-cli/src/neuro_cli/blob_storage.py
```diff
@@ -408,7 +427,7 @@ async def cp(
                 param_type="argument", param_hint='"SOURCES..."'
             )
         sources = *sources, destination
-        target_dir = parse_blob_or_file_resource(target_directory, root)
+        target_dir = target_directory
```
asvetlov, line 430: Removing `target_dir` and `dst` variables for the sake of reusing already existing `target_directory` and `destination` makes the code a bit more readable and understandable.

asvetlov (approved):
LGTM in general, please consider the fixing mentioned minor comments.

asvetlov (commented):
Please merge :)

PR #2423, opened 2021-11-24
Use neuro-admin-client to access admin api

romasku:


neuro-sdk/tests/test_public_names.py
```diff
@@ -14,9 +14,11 @@ def test_module_for_public_names() -> None:
...
         if hasattr(obj, "__module__"):
-            # objects from typing are public type hint aliases,
-            # e.g. Callable[...]
             assert obj.__module__ in (
                 "neuro_sdk",
+                # objects from typing are public type hint aliases,
+                # e.g. Callable[...]
                 "typing",
+                # We re-export entities from admin client
+                "neuro_admin_client.entities",
```
romasku, line 23: I'm completely not sure about it. Is there any better solution? I see the following options, but I can't say that they are better:
- Do not re-export at all and require users to import from `neuro_admin_client.entities`. As for me, this is just ugly.
- Wrap all entities from `neuro_admin_client` into SDK descendants. It will require a lot of boilerplate code that just maps one dataclass to corresponding from SDK. And we will need to wrap **all** methods of client to do this transform.

Any better suggestions?
Also, why is it important to have "neuro_sdk" as `__module__`?

asvetlov: I agree with the test relaxing.
The main reason for the module rewriting is the public interface.
If we document that *canonical* name for `JobDescription` is `neuro_sdk.JobDescription` but `repr(job)` points on `neuro_sdk._jobs.JobDescription` -- it looks weird.
Sphinx autodocs plugin fails if `__module__` is not tuned well.
Also, autocompletion tools might use `__module__` for suggestions, it actually depends on the used tool and its version.

asvetlov (approved):
LGTM.
Please make `neuro-admin-client` release first.

PR #2436, opened 2021-11-26
Merge 'config show-quota' into 'config show' CLI command

asvetlov:


YevheniiSemendiak:
Please, ping also @jane-gorlova regarding such changes in CLI, since our docs quickly become outdated.

PR #2437, opened 2021-11-26
Add cli-command-checks and fix new lint rules

romasku:


romasku:
Maybe you have an idea of some other checkers we can add? Is there some spelling check available to integrate it easily?. Also, we are not consistent in writing "." after the short doc, we can also add a check for this. WDYT?

asvetlov (approved):
LGTM

PR #2481, opened 2021-12-16
Support orgs in bucket urls

romasku:
Unfortunately, it's impossible to parse url like `blob://cluster/part1/part2/part3/part4`. I can be any of the following:
- `part1` is username, `part2` is bucket name, `part3/part4` is path to blob.
- `part1/part2` is username with slash, `part3` is bucket name, `part4` is path to blob.
- `part1` is org, `part2` is username, `part3` is bucket name,  `part4` is path to blob.
(and some even more options). 
So I have to fetch all buckets in clusters to properly parse url.


asvetlov (commented):
Sounds ugly, sorry.
Let's discuss offline

PR #2496, opened 2021-12-23
List roles

romasku:
Blocked by https://github.com/neuro-inc/platform-auth/pull/692
Closes https://github.com/neuro-inc/platform-client-python/issues/2395

### DIFF
--- a/neuro-sdk/src/neuro_sdk/_users.py
+++ b/neuro-sdk/src/neuro_sdk/_users.py
@@ -102,6 +102,16 @@ class Users(metaclass=NoPublicConstructor):
             ret.append(Share(item["user"], Permission(uri, action)))
         return ret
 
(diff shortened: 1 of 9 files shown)


neuro-sdk/src/neuro_sdk/_users.py
```diff
@@ -102,6 +102,16 @@ async def get_shares(
             ret.append(Share(item["user"], Permission(uri, action)))
         return ret
 
+    async def get_subroles(
```
asvetlov, line 105: Would you write a test for this method?

romasku: Sorry, forgot about it. Done.

asvetlov (approved):

PR #2519, opened 2022-01-10
Drop test clusters using platform-admin list

romasku:


asvetlov (approved):

neuro-cli/src/neuro_cli/admin.py
```diff
@@ -53,6 +53,18 @@ async def get_clusters(root: Root) -> None:
         root.print(fmt(clusters.values()))
 
 
+@command(hidden=True)
```
asvetlov, line 56: Is there reason to make the command hidden?

romasku: This command is mostly the same as `get-clusters` and can return different data only for clusters that failed to set up properly on the platform-config side. It is very low-level and can be useful to debug very strange situations. So I prefer to hide it from normal users as it is hard to explain why we need this command.

asvetlov: Understood

PR #2560, opened 2022-01-26
support org storage size parameter

zubenkoivan:


neuro-cli/tests/e2e/test_e2e_admin.py
```diff
@@ -376,6 +376,7 @@ def test_remove_org_user_does_not_exist(helper: Helper, tmp_test_org: str) -> No
 
 
 @pytest.mark.e2e
+@pytest.mark.skip
```
zubenkoivan, line 379: Skip cluster-org tests for now as they require infrastructure provisioning until some kind of fake cluster is implemented.

asvetlov (approved):

PR #2561, opened 2022-01-26
Send "org: 'null'" for jobs with no org

romasku:
Closes https://github.com/neuro-inc/neuro-cli/issues/2562


romasku:
This was fixed on the server side: https://github.com/neuro-inc/platform-api/commit/e00518a2946958b49ccaf7934b50047be662a7c8

PR #2573, opened 2022-01-28
Fix output of neuro admin update-org-cluster

romasku:
Closes https://github.com/neuro-inc/neuro-cli/issues/2559


asvetlov (approved):
Would be nice to have a test for `OrgClusterFormatter`, as we have for other formatters already.

PR #2579, opened 2022-01-31
Fix resolving job URIs containing organization name

### LINKED ISSUE #2572: Not working URIs for jobs in organization
`neuro status` and any other command which takes a job URI does not work with a job ran in organization.
```
$ neuro status job:job-3cfa7099-2cda-42ca-9afb-9b2582b4b2d6
ERROR: Illegal argument(s) ({0: DataError('value should be None'), 1: DataError('String is longer than 40 characters')})
$ neuro status job://default/test-org/serhiystorchaka/job-3cfa7099-2cda-42ca-9afb-9b2582b4b2d6
ERROR: Illegal argument(s) ({0: DataError('value should be None'), 1: DataError('String is longer than 40 characters')})
$ neuro status job:test-job2
ERROR: Illegal argument(s) ({0: DataError('value should be None'), 1: DataError('does not match pattern \\A[a-z](?:-?[a-z0-9])*\\Z')})
$ neuro status job://default/test-org/serhiystorchaka/test-job2
ERROR: Illegal argument(s) ({0: DataError('value should be None'), 1: DataError('does not match pattern \\A[a-z](?:-?[a-z0-9])*\\Z')})
```
It is because the full URI `job://{cluster}/{org}/{user}/{jobid}` is parsed as `cluster={cluster}`, `owner={org}` and `name={user}/{jobid}`.

serhiy-storchaka:
Closes #2572.

### DIFF
--- a/CHANGELOG.D/2572.bugfix
+++ b/CHANGELOG.D/2572.bugfix
@@ -1,1 +1,1 @@
-Fix resolving job URIs containing organization name.
+Fix resolving job URIs containing organization name and/or service account name.
--- a/CHANGELOG.D/2572.bugfix
+++ b/CHANGELOG.D/2572.bugfix
@@ -1,0 +1,1 @@
+Fix resolving job URIs containing organization name.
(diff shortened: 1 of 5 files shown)


romasku (approved):
Looks good to me, I like the solution.

PR #2585, opened 2022-02-07
support disk usage per organization

zubenkoivan:


romasku (approved):

romasku:
Looks good. 
As separate PR, we should probably allow users to manually specify the path in `neuro storage df`:
`neuro storage df storage://cluster/another-org/...`
This way cluster manager will not have to jump between orgs just to check usage.

PR #2699, opened 2022-05-23
Use sentinel as default value for org_name in sdk

romasku:
Closes https://github.com/neuro-inc/neuro-cli/issues/2677

### DIFF
--- a/neuro-sdk/src/neuro_sdk/_buckets.py
+++ b/neuro-sdk/src/neuro_sdk/_buckets.py
@@ -48,10 +48,10 @@ from ._parser import Parser
 from ._rewrite import rewrite_module
 from ._url_utils import _extract_path, normalize_local_path_uri
 from ._utils import (
+    ORG_NAME_SENTINEL,
     NoPublicConstructor,
     OrgNameSentinel,
     asyncgeneratorcontextmanager,
-    org_name_sentinel,
 )
 
 
@@ -233,7 +233,7 @@ class Buckets(metaclass=NoPublicConstructor):
         self,
         name: Optional[str] = None,
         cluster_name: Optional[str] = None,
-        org_name: Union[Optional[str], OrgNameSentinel] = org_name_sentinel,
+        org_name: Union[Optional[str], OrgNameSentinel] = ORG_NAME_SENTINEL,
     ) -> Bucket:
         url = self._get_buckets_url(cluster_name)
         auth = await self._config._api_auth()
@@ -254,7 +254,7 @@ class Buckets(metaclass=NoPublicConstructor):
         credentials: Mapping[str, str],
         name: Optional[str] = None,
         cluster_name: Optional[str] = None,
-        org_name: Union[Optional[str], OrgNameSentinel] = org_name_sentinel,
+        org_name: Union[Optional[str], OrgNameSentinel] = ORG_NAME_SENTINEL,
     ) -> Bucket:
         url = self._get_buckets_url(cluster_name) / "import" / "external"
         auth = await self._config._api_auth()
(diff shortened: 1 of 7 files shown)


neuro-sdk/src/neuro_sdk/_utils.py
```diff
@@ -27,6 +27,14 @@
 from ._errors import ConfigError
 from ._rewrite import rewrite_module
 
+
+class OrgNameSentinel:
+    pass
+
+
+org_name_sentinel = OrgNameSentinel()
```
asvetlov, line 35: Make the constant uppercase please

asvetlov (changes requested):

asvetlov (approved):

neuro-sdk/src/neuro_sdk/_buckets.py
```diff
@@ -228,13 +233,15 @@ async def create(
         self,
         name: Optional[str] = None,
         cluster_name: Optional[str] = None,
-        org_name: Optional[str] = None,
+        org_name: Union[Optional[str], OrgNameSentinel] = ORG_NAME_SENTINEL,
```
serhiy-storchaka, line 236: Should the documentation be updated?

Also, you can use `Union[str, None, Literal[ORG_NAME_SENTINEL]]` as a type annotation.

neuro-sdk/src/neuro_sdk/_buckets.py
```diff
@@ -228,13 +233,15 @@ async def create(
...
     ) -> Bucket:
         url = self._get_buckets_url(cluster_name)
         auth = await self._config._api_auth()
         data = {
             "name": name,
-            "org_name": org_name or self._config.org_name,
+            "org_name": org_name
+            if not isinstance(org_name, OrgNameSentinel)
```
serhiy-storchaka, line 243: Or maybe just `if org_name is not ORG_NAME_SENTINEL`?

neuro-sdk/src/neuro_sdk/_utils.py
```diff
@@ -27,6 +27,14 @@
 from ._errors import ConfigError
 from ._rewrite import rewrite_module
 
+
+class OrgNameSentinel:
+    pass
```
serhiy-storchaka, line 32: If add a special class for a singleton, I would add `__repr__`.

serhiy-storchaka (approved):
LGTM. Added few minor questions/comments.

neuro-sdk/src/neuro_sdk/_buckets.py
romasku: Hmm, probably I missing something. `Literal` generic can only work with literal types (like ints, strings, bools, None) and enums. But `OrgNameSentinel` is just ordinary class. Do you suggest it to replace with an enum?

neuro-sdk/src/neuro_sdk/_buckets.py
romasku: In the current code, mypy cannot deduce that `OrgNameSentinel` is a singleton, so it can only narrow the type after `isinstance` check. Maybe `Literal` will can solve this problem.

neuro-sdk/src/neuro_sdk/_utils.py
romasku: Sorry, I'm not getting your point. What kind of repr do you suggest? When will it be used?

neuro-sdk/src/neuro_sdk/_buckets.py
asvetlov: Correct. There is a draft PEP for singletons, it is not implemented AFAIK

neuro-sdk/src/neuro_sdk/_utils.py
asvetlov: `__repr__` method of `OrgNameSentinel` would be good to have but it is not mandatory.

PR #2711, opened 2022-05-27
Displaay some additional information for imported buckets

### LINKED ISSUE #2706: `blob statbucket` should also display bucket endpoint URI
#### Is your feature request related to a problem? Please describe.
If the bucket was imported some time ago, it becomes hard to remember what is the origin of this bucket.

#### Describe the solution you'd like
Having an endpoint URI included in the command `neuro blob statbucket` result will help with that.

#### Additional context
Neuro Platform Client 22.1.3

serhiy-storchaka:
`neuro blob statbucket` now displays some non-secret credential
information for imported buckets.

For AWS: "External name", "External endpoint", "External region name".
For Azure: "External name", "External endpoint".
For GCP: "External name".

Closes #2706.

### DIFF
--- a/CHANGELOG.D/2706.feature
+++ b/CHANGELOG.D/2706.feature
@@ -1,0 +1,6 @@
+`neuro blob statbucket` now displays some additional information
+for imported buckets.
+
+* For AWS: "External name", "External endpoint", "External region name".
+* For Azure: "External name", "External endpoint".
+* For GCP: "External name".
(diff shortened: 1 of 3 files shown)


serhiy-storchaka:
Do we need a special option for this, or display it by default? On my computer it increases the time of `neuro blob statbucket` from 2.5-3 seconds to 4-4.5 seconds.

Should it display also more secret data?

romasku:
I think it should be enabled by default. This command will be only used when the user wants to "inspect" all details about a given bucket, so it's OK to add some delay but to show everything we have.
Short info is already available in the `ls` command.

romasku (approved):

asvetlov:
Ok

PR #2725, opened 2022-06-08
support users without clusters

zubenkoivan:


asvetlov (approved):

neuro-sdk/src/neuro_sdk/_config.py
```diff
@@ -59,7 +59,7 @@ class _ConfigData:
     url: URL
     admin_url: Optional[URL]
     version: str
-    cluster_name: str
+    cluster_name: Optional[str]
```
asvetlov, line 62: Please reflect the type change in https://github.com/neuro-inc/neuro-cli/blob/master/neuro-sdk/docs/config_reference.rst#L50

zubenkoivan: This type didn't change. It is internal dataclass.

PR #2728, opened 2022-06-09
Do config reload in neuro config show

romasku:


asvetlov (approved):
Add a changelog please

PR #2733, opened 2022-06-15
support neuro sdk job status method by users without clusters

zubenkoivan:


neuro-sdk/src/neuro_sdk/_jobs.py
```diff
@@ -873,9 +873,11 @@ def _http_port_from_api(data: Dict[str, Any]) -> HTTPPort:
     )
 
 
-def _container_from_api(data: Dict[str, Any], parse: Parser) -> Container:
+def _container_from_api(
+    data: Dict[str, Any], cluster_name: str, parse: Parser
```
asvetlov, line 877:
```suggestion
    data: Dict[str, Any], cluster_name: str, parse: Parser,
```

asvetlov (approved):

PR #2748, opened 2022-06-27
Fix CLI code and tests for user with org

### LINKED ISSUE #2747: Many tests are failed for user with org
For example, the test expects string `image://default/neuro-cli-e2e/e2e-cli-cbe71fec-3786-44bb-8c17-377f0042937f-date202206271144-date-pushed:e3f05c14-f711-4c83-9aed-4251c3727c9c`, but the output contains string `image://default/e2e-testing-016c101540759229dc1f/neuro-cli-e2e/e2e-cli-cbe71fec-3786-44bb-8c17-377f0042937f-date202206271144-date-pushed:e3f05c14-f711-4c83-9aed-4251c3727c9c`.

serhiy-storchaka:
* Disks, buckets and secrets are now created with the current organization instead of no organization if `--org` is not explicitly specified. (#2756)
* Fix formatting URIs in short form for users with organization. (#2749)
* Fix E2E tests for users with organization. (#2747)


asvetlov:
@serhiy-storchaka please fix the conflict
Maybe @romasku applied the similar fix to master already?

serhiy-storchaka:
Already fixed the conflict locally, but do not want to push the branch until fix other tests, because it will trigger CI testing which will only increase a load. The conflict was due to incompatibility with new `neuro-admin-client`, which @romasku fixed in other way.

https://github.com/neuro-inc/neuro-cli/pull/2751 fixed unintentional running tests in organization which exposed problems. This PR fixes that problems.

asvetlov (approved):

PR #2791, opened 2022-07-19
integrate neuro-config-client

zubenkoivan:


neuro-sdk/src/neuro_sdk/_clusters.py
```diff
@@ -0,0 +1,117 @@
...
+    "_StorageOptions",
+    "_TPUPreset",
+    "_VCDCloudProvider",
+    "_VCDCloudProviderOptions",
+    "_VCDStorage",
+]
+
+
+@rewrite_module
+class _Clusters(ConfigClientBase, metaclass=NoPublicConstructor):
```
asvetlov, line 82: Are you sure that the inheritance from `ConfigClientBase` is the best solution here?
Do we need to expose all `ConfigClientBase` methods in SDK?
I would embed it rather than inherit and provide a simple subset of required API methods.

zubenkoivan: I just used the same approach we have in for `_Admin`

asvetlov (approved):

PR #2903, opened 2023-02-08
Support energy schedule

### LINKED ISSUE #2888: Support specifying energy schedule on job submit
We should also report the chosen energy schedule for existing jobs on `status`

YevheniiSemendiak:
closes  #2888

I didn't change the internal config, instead used neuro-config-client to fetch a list of energy schedules within the cluster.

- Potentially, the table format of energy schedules should be changed.
take a look at [example](neuro-cli/tests/unit/formatters/ascii/TestConfigFormatter.test_output_for_energy_schedules_0.ref)
- Should I also include cluster timezone to the `neuro config show --energy` report?

I also need to test it with the live schedule on dev, but please take a look at the initial implementation.


neuro-cli/src/neuro_cli/config.py
```diff
@@ -48,7 +53,10 @@ async def show(root: Root) -> None:
             jobs_capacity = {}
     quota = await root.client.users.get_quota()
     org_quota = await root.client.users.get_org_quota()
-    root.print(fmt(root.client.config, jobs_capacity, quota, org_quota))
+    config_cluster = None
+    if energy:
+        config_cluster = await root.client._clusters.get_cluster(cluster_name)
```
dalazx, line 58: where do we read the cluster information from?

I see that we also render price_kwh in that output table. `/api/v1/config` does not output neither price nor co2.

neuro-cli/src/neuro_cli/job.py
```diff
@@ -1467,9 +1479,11 @@ def _job_to_cli_args(job: JobDescription) -> List[str]:
         res += ["--privileged"]
     if job.priority != JobPriority.NORMAL:
         res += ["--priority", job.priority.name.lower()]
+    if job.energy_schedule_name:
+        res += ["--energy-schedule", job.energy_schedule_name]
     res += [str(job.container.image)]
     if job.container.command:
-        res += [job.container.command]
+        res += ["--", job.container.command]
```
dalazx, line 1486: is this needed? maybe we should have a dedicated PR for this change?

dalazx: or this is super minor and won't possibly break anything?

neuro-cli/tests/unit/formatters/ascii/TestConfigFormatter.test_output_for_energy_schedules_0.ref
```diff
@@ -0,0 +1,41 @@
...
+ gpu-small      7   32.2 GB        ×               ×           1 x nvidia-tesla-k80    10               
+ gpu-large      7   64.4 GB        ×               ×           1 x nvidia-tesla-v100   10               
+ cpu-small      7    2.1 GB        ×               ×                                   10               
+ cpu-large      7   15.0 GB        ×               ×                                   10               
+Cluster energy parameters
+CO2 emmitions (g/kWh): 40.4
+Energy schedules                                                        
+ Name      Price (kW/h)   Weekday   Period start time   Period end time 
+━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+ DEFAULT        0            1          00:00:00           23:59:59
```
dalazx, line 21: could we drop seconds please?

dalazx (commented):
great job!

neuro-cli/src/neuro_cli/job.py
YevheniiSemendiak: Haven't found that we use it elsewhere except our CLI (here only in generate_run_command).

This is mostly personal, I often use it to debug and the deprecation warning is annoying, decided to make this small fix in place :) Apologize, next time I'll open a dedicated PR for such things.

neuro-cli/src/neuro_cli/config.py
YevheniiSemendiak: This is taken from https://github.com/neuro-inc/neuro-config-client, see trace ([1](https://github.com/neuro-inc/neuro-cli/blob/ys/support-energy-schedule/neuro-sdk/src/neuro_sdk/_client.py#L30), [2](https://github.com/neuro-inc/neuro-cli/blob/ys/support-energy-schedule/neuro-sdk/src/neuro_sdk/_clusters.py#L126))

As far as I could see, we were transitioning between deprecating and removing neuro-config-client-related functionality within SDK and using dedicated neuro-config-client for this.

dalazx (approved):
👍

YevheniiSemendiak:
@dalazx , thx for the reviews!
One last note: tests for py.37 are failing bc of bumped pre-commit hook (they've dropped support for 3.7 recently). I do not see the reason for installing previous version for pre-config for 3.7 tests, since we will drop 3.7 support for CLI too (it's EOL is scheduled for 2023-06-27).
I'm also awaiting for energy configs in dev to test this thing (continuing in [Slack](https://neuromation.slack.com/archives/CE76BT03G/p1676210983400389?thread_ts=1676028361.160839&cid=CE76BT03G)).

PR #2933, opened 2023-03-21
Fix human-readable datetime formatter to correctly report jobs lifespan ending in future

YevheniiSemendiak:
Closes https://neuromation.slack.com/archives/CE9SZE5B3/p1679391955072579


neuro-cli/src/neuro_cli/formatters/utils.py
```diff
@@ -107,15 +107,21 @@ def format_datetime_human(
     assert when.tzinfo is not None
     delta = datetime.datetime.now(datetime.timezone.utc) - when
     if delta < datetime.timedelta(days=1):
+        preffix = ""
```
andriihomiak, line 110: it probably should be `prefix`, not `preffix`

andriihomiak (changes requested):
Looks good, just one minor typo

PR #2958, opened 2023-04-18
Use project and org name while logging in from passed config

YevheniiSemendiak:
closes https://github.com/neuro-inc/neuro-cli/issues/2947

### DIFF
--- a/neuro-sdk/src/neuro_sdk/_config_factory.py
+++ b/neuro-sdk/src/neuro_sdk/_config_factory.py
@@ -235,12 +235,17 @@ class Factory:
             token = data["token"]
             cluster = data["cluster"]
             url = URL(data["url"])
+            org_name = data.get("org_name")
+            project_name = data.get("project_name")
         except (ValueError, KeyError):
             raise ConfigError(f"Data in passed config is malformed: {config_data}")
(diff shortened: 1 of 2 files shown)


neuro-sdk/src/neuro_sdk/_config_factory.py
```diff
@@ -235,12 +235,17 @@ async def login_with_passed_config(
...
             url = URL(data["url"])
+            org_name = data.get("org_name")
+            project_name = data.get("project_name")
         except (ValueError, KeyError):
             raise ConfigError(f"Data in passed config is malformed: {config_data}")
         await self.login_with_token(token, url=url, timeout=timeout)
         client = await self.get(timeout=timeout)
 
         await client.config.switch_cluster(cluster)
+        await client.config.switch_org(org_name)
```
dalazx, line 246: curious whether this line should also be conditional.

dalazx (approved):
looks good

neuro-sdk/src/neuro_sdk/_config_factory.py
YevheniiSemendiak: yes, switch_org accepts none and sets local config w.o. org

PR #2959, opened 2023-04-21
support projects in storage

### LINKED ISSUE #2949: Support projects in platform storage
Part of https://github.com/neuro-inc/no-code/issues/77
We want to be able to parameterize the project name for cp/glob/ls/mkdir/mv/rm/tree commands.

zubenkoivan:
closes #2949


neuro-sdk/src/neuro_sdk/_jobs.py
```diff
@@ -1131,8 +1131,9 @@ def _job_telemetry_from_api(value: Dict[str, Any]) -> JobTelemetry:
 
 
 def _volume_to_api(volume: Volume, config: Config) -> Dict[str, Any]:
+    assert config.project_name, "Project is not configured"
```
dalazx, line 1134: minor. maybe we should also raise RuntimeError here? AssertionError might be disabled if the interpreter is run with -p or something.

dalazx (approved):
lgtm

PR #2960, opened 2023-04-23
Add support of project in disks

YevheniiSemendiak:
closes https://github.com/neuro-inc/neuro-cli/issues/2951


neuro-cli/src/neuro_cli/disks.py
```diff
@@ -165,17 +187,23 @@ async def create(
...
+) -> None:
     """
     Get disk DISK_ID.
     """
-    disk_id = await resolve_disk(disk, client=root.client, cluster_name=cluster)
-    disk_obj = await root.client.disks.get(disk_id, cluster_name=cluster)
+    if isinstance(disk, URL):
+        disk_str = disk.parts[-1]
+    else:
+        disk_str = disk
```
dalazx, line 199: could you please explain why we dropped `resolve_disk` here?

neuro-cli/src/neuro_cli/formatters/utils.py
```diff
@@ -27,7 +30,11 @@ def formatter(uri: URL) -> str:
                     owner, _, rest = path.partition("/")
                 else:
                     owner = owner_or_org
-                if owner == username:
+                # TODO (Y.S.): remove this check and use project_name parameter
```
dalazx, line 33: should we assume right away that there is no owner in URIs?

neuro-sdk/src/neuro_sdk/_disks.py
```diff
@@ -85,10 +87,22 @@ def _get_disks_url(self, cluster_name: Optional[str]) -> URL:
...
+        org_name: Optional[str] = None,
+        project_name: Optional[str] = None,
+    ) -> AsyncIterator[Disk]:
         url = self._get_disks_url(cluster_name)
+        params = {}
+        org = org_name or self._config.org_name
+        if org:
+            params["org_name"] = org
+        if project_name:
+            params["project_name"] = project_name
```
dalazx, line 102: this might seem a bit inconsistent in terms of other methods where we fall back to `_config` if `project_name` is missing.

in general, using `cluster_name` as is but falling back to `org_name` and `project_name` seems a bit off, but we probably don't want to dive into this now.

neuro-cli/src/neuro_cli/disks.py
YevheniiSemendiak: Good question, originally the intent was to simplify this logic. When I started to explain possible cases, I find out that my reasoning was wrong and we still need this method for cases when we have disk URLs in full form or with the owner.
Reverted in [c381a1e](https://github.com/neuro-inc/neuro-cli/pull/2960/commits/c381a1ecf85a8bc5423310dac2449a8259a72415).

neuro-sdk/src/neuro_sdk/_disks.py
YevheniiSemendiak: Yes, let's agree on the behavior for those operations and create a corresponding issue to tackle it separately, wdyt?

Here I followed the reasoning: creating resources always requires `project_name` so we fall back. The listing does not require it, so we are safe to list everything available for the user in favor of UX (unless we filter results). As for the update, if the resources could not be identified (referred by name, not ID), we fall back.

dalazx (approved):
lgtm

neuro-sdk/src/neuro_sdk/_disks.py
YevheniiSemendiak: added https://github.com/neuro-inc/neuro-cli/issues/2962

PR #2963, opened 2023-04-25
Pin yarl version

YevheniiSemendiak:
https://github.com/aio-libs/yarl/issues/854#issuecomment-1520182983


YevheniiSemendiak:
tested and suppressed by https://github.com/neuro-inc/neuro-cli/pull/2970

PR #2970, opened 2023-05-05
support projects in blob commands

### LINKED ISSUE #2954: Support project in platform buckets
Part of https://github.com/neuro-inc/no-code/issues/77
We want to be able to parameterize the project name for buckets, their content, and credentials.

zubenkoivan:
closes #2954

### DIFF
--- a/neuro-cli/src/neuro_cli/blob_storage.py
+++ b/neuro-cli/src/neuro_cli/blob_storage.py
@@ -22,6 +22,7 @@ from neuro_cli.click_types import (
     BUCKET_NAME,
     CLUSTER,
     ORG,
+    PROJECT,
     PlatformURIType,
 )
 from neuro_cli.formatters.bucket_credentials import (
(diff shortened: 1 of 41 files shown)


neuro-cli/src/neuro_cli/click_types.py
```diff
@@ -1226,3 +1248,4 @@ async def async_shell_complete(
 def setup_shell_completion() -> None:
     add_completion_class(NewZshComplete)
     add_completion_class(NewBashComplete)
+    add_completion_class(NewBashComplete)
```
dalazx, line 1251: probably a duplicate line.

dalazx (approved):
one minor comment. lgtm.

neuro-cli/src/neuro_cli/blob_storage.py
```diff
@@ -140,6 +164,7 @@ async def mkbucket(
     name: Optional[str] = None,
     cluster: Optional[str] = None,
     org: Optional[str] = None,
+    project: Optional[str] = None,
```
YevheniiSemendiak, line 167: minor: do we forbid the creation of buckets with the same name in different projects?
currently, I cannot do it and it goes from the API.

YevheniiSemendiak (approved):
Great job! Thank you :muscle:

neuro-cli/src/neuro_cli/blob_storage.py
zubenkoivan: Should be possible. When did you test this last time?

zubenkoivan: I think I found a bug in buckets API, will work on this, thank you

PR #2995, opened 2023-06-01
align all ls command options

zubenkoivan:


neuro-cli/src/neuro_cli/image.py
```diff
@@ -137,9 +149,21 @@ async def ls(
     with root.status("Fetching images"):
         images = await root.client.images.list(cluster_name=cluster)
 
-    if project:
-        projects = set(project)
-        images = [image for image in images if image.project_name in projects]
+    if all_orgs:
```
YevheniiSemendiak, line 152: Our image name parser works incorrect - it threats org names as project names and the underlying project name as a part of image name.
AFAIU, this affects only the representation (and filtering here), so we could fix it later / in a separate PR.

example:
(current config: testproj + no-org, testorg is an org with two projects, yevheniisemendiak and testproj and tp4 is a project w.o. org the current user has access to)
image

neuro-sdk/src/neuro_sdk/_jobs.py
```diff
@@ -518,6 +519,8 @@ async def list(
             params.add("being_dropped", str(_being_dropped))
         if _logs_removed is not None:
             params.add("logs_removed", str(_logs_removed))
+        for org_name in org_names:
+            params.add("org_name", org_name or "NO_ORG")
```
YevheniiSemendiak, line 523: As I could see, we do not support `NO_ORG` filter on the API side yet.

YevheniiSemendiak (commented):
Great job!
Some suggestions here. We might address most of them in a separate PRs, but I would rather suggest resolving default filtering of `neuro job ps` (which is broken atm).

neuro-sdk/src/neuro_sdk/_jobs.py
YevheniiSemendiak: submited https://github.com/neuro-inc/platform-api/issues/2096

neuro-cli/src/neuro_cli/secrets.py
```diff
@@ -31,44 +32,57 @@ def secret() -> None:
...
         )
 
+    if all_orgs:
+        org_name = ORG_NAME_SENTINEL
+    else:
+        org_name = parse_org_name(org, root)  # type: ignore
+
+    if all_projects:
+        project_name = None
+    else:
```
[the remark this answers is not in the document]
zubenkoivan: I doubt that they are conflicting. They just select orgs/projects set to show, and `--all-orgs` covers `--org` option.

neuro-sdk/src/neuro_sdk/_jobs.py
zubenkoivan: thanks, have already started working on it

neuro-cli/src/neuro_cli/image.py
zubenkoivan: I will do it in a separate pr. AFAIU it was broken when we introduced orgs, parser doesn't recognize non current orgs and treats them as project.

neuro-cli/src/neuro_cli/secrets.py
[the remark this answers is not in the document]
YevheniiSemendiak: True, the code works as expected, I just refer to UX. This is minor

YevheniiSemendiak (approved):
LGTM!

PR #2998, opened 2023-06-09
fix resolve_disk method

zubenkoivan:
Currently disks are resolved by (cluster_name,owner). Switched to resolve by (cluster_name,org_name,project_name)


YevheniiSemendiak (approved):
Looks good! Thank you for fixing it! :+1:

PR #3001, opened 2023-07-03
chore: Prepare for 23.7.0 release

andriihomiak:


requirements/base.txt
```diff
@@ -6,7 +6,7 @@ aiohttp==3.8.4
 azure-core==1.26.3
 azure-storage-blob==12.14.1
 # upgrade setup.py after pinning certtifi to new version
-certifi==2022.12.7
+certifi==2023.5.7
```
YevheniiSemendiak, line 9: do we need this update? might cause some conflicts, but not sure

andriihomiak: You are right, lets skip it for now and address it in the future update

PR #3137, opened 2024-11-29
END-486: new cmds for org/user balance

parikls:


asvetlov (approved):
Looks great, thank you!

Please add a file `./CHANGELOG.D/3137.feature` with new feature description. A sentence or two in a free form should be enough, Markdown formatting is supported if needed.

CHANGELOG.D/3137.feature
```diff
@@ -0,0 +1,12 @@
+Balance is no longer stored on a cluster level, and was moved to an organization level, e.g.,
+to an org itself, and to an org users, instead of a cluster / cluster users.
+
+New commands:
+  - `set_org_defaults` - allows to set an organization default values, such as a default user credits
```
asvetlov, line 5:
```suggestion
  - `apolo admin set-org-defaults` - allows to set an organization default values, such as a default user credits
```
`click` replaces underscores with dashes.
The full command path is more informative from end user point of view maybe, but the final decision is up to you.

The same is true for other mentioned commands.

CHANGELOG.D/3137.feature
```diff
@@ -0,0 +1,12 @@
...
+
+Existing commands changes:
+  - `add_cluster_user` cmd is no longer accepting a `credits` argument.
+  - `set_user_credits` cmd is now expecting an org name instead of a cluster name.
+  - `add_user_credits` cmd is now expecting an org name instead of a cluster name.
+  - `set_org_cluster_credits` is now deprecated in a favor of a `set_org_credits`.
```
asvetlov, line 11: `set_org_cluster_credits` is not a thing. Do you mean `set_org_cluster_quota`?
I don't think that we need keeping deprecated admin commands, we can break backward compatibility in this command group without any problem.

Let's just drop `set_org_cluster_quota` if we don't need it. If you insist on keeping the command please add `deprecated=True` argument to `@command()` decorator.

CHANGELOG.D/3137.feature
```diff
@@ -0,0 +1,12 @@
...
+  - `add_org_cluster_credits` is not deprecated in a favor of an `add_org_credits`.
```
asvetlov, line 12: Sorry, I cannot find `add_org_cluster_credits` in `admin.py`

asvetlov (commented):
I have a few minor notes.

Also, it would be nice of changelog record will reflect the formatters change also.

CHANGELOG.D/3137.feature
parikls: @asvetlov it was removed, I've updated the changelog text.

CHANGELOG.D/3137.feature
parikls: @asvetlov it was removed, I've updated the changelog text.

CHANGELOG.D/3137.feature
parikls: @asvetlov changed to a full commands everywhere and used a dashes. Thanks!

asvetlov (approved):
Great, thanks!

PR #3142, opened 2024-12-05
MLO-46: Expose apps config to users

YevheniiSemendiak:
We don't need it in CLI, only in SDK at the moment.

### DIFF
--- a/apolo-sdk/src/apolo_sdk/_server_cfg.py
+++ b/apolo-sdk/src/apolo_sdk/_server_cfg.py
@@ -85,7 +85,7 @@ class Project:
 @rewrite_module
 @dataclass(frozen=True)
 class AppsConfig:
-    apps_hostname_templates: Sequence[str] = ()
+    hostname_templates: Sequence[str] = ()
 
 
 @rewrite_module
@@ -188,11 +188,12 @@ def _parse_cluster_config(payload: Dict[str, Any]) -> Cluster:
         orgs = [org if org is not None else "NO_ORG" for org in orgs]
 
     apps_payload = payload.get("apps", {})
-    apps_config = AppsConfig()
     if apps_payload:
         apps_config = AppsConfig(
-            apps_hostname_templates=apps_payload.get("apps_hostname_templates", [])
+            hostname_templates=apps_payload.get("apps_hostname_templates", [])
         )
+    else:
+        apps_config = AppsConfig()
 
     cluster_config = Cluster(
         name=payload["name"],
--- a/apolo-sdk/src/apolo_sdk/_server_cfg.py
+++ b/apolo-sdk/src/apolo_sdk/_server_cfg.py
@@ -84,6 +84,12 @@ class Project:
 
 @rewrite_module
 @dataclass(frozen=True)
+class AppsConfig:
+    apps_hostname_templates: Sequence[str] = ()
+
+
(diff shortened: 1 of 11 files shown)


apolo-sdk/src/apolo_sdk/_server_cfg.py
```diff
@@ -82,6 +82,12 @@ def key(self) -> Key:
         )
 
 
+@rewrite_module
+@dataclass(frozen=True)
+class AppsConfig:
+    apps_hostname_templates: Sequence[str] = ()
```
asvetlov, line 88: I'm curious what is the meaning of this list?
What is the format for the list item?

apolo-sdk/src/apolo_sdk/_server_cfg.py
```diff
@@ -179,6 +186,14 @@ def _parse_cluster_config(payload: Dict[str, Any]) -> Cluster:
         orgs = ["NO_ORG"]
     else:
         orgs = [org if org is not None else "NO_ORG" for org in orgs]
+
+    apps_payload = payload.get("apps", {})
+    apps_config = AppsConfig()
+    if apps_payload:
+        apps_config = AppsConfig(
+            apps_hostname_templates=apps_payload.get("apps_hostname_templates", [])
+        )
```
asvetlov, line 195: Let's avoid calling `AppsConfig` constructor twice.
```suggestion
    if apps_payload:
        apps_config = AppsConfig(
            apps_hostname_templates=apps_payload.get("apps_hostname_templates", [])
        )
    else:
        apps_config = AppsConfig()
```

asvetlov (commented):
Should we display this config in commands like `apolo config show`?

Also, the documentation should be updated to reflect SDK changes.

apolo-sdk/src/apolo_sdk/_server_cfg.py
asvetlov, line 88: Could we use a shorter name, `apps_hostname_templates` is too clumsy.

YevheniiSemendiak: This is somewhat similar to `job_hostname_template`. This is a string with a placeholder `"{app_name}.apps.novoserve.org.neu.ro."`.  Also, Ivan mentioned, we might have a list of domain names for a single cluster in a future, therefore, it's a list of such placeholders.

End customers will not use it directly at the moment, therefore, I didn't add it to CLI itself.
We are going to use it in our apps deployment workflow, see https://linear.app/apolo/project/apps-support-various-base-domains-3e84552057e0/overview

apolo-sdk/src/apolo_sdk/_server_cfg.py
YevheniiSemendiak: applied, thank you!

apolo-sdk/src/apolo_sdk/_server_cfg.py
asvetlov: ok

asvetlov (commented):
Along with documentation update please create `./CHANGELOG.D/3142.feature` file

YevheniiSemendiak:
@asvetlov added docs, changelog. Please, take a glance once more

asvetlov (approved):
All good, thanks!

PR #3149, opened 2024-12-16
remove cluster storage options

zubenkoivan:


apolo-sdk/src/apolo_sdk/_config_factory.py
```diff
@@ -57,7 +57,7 @@ async def __make_session(
     from . import __version__
 
     ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
-    ssl_context.load_verify_locations(capath=certifi.where())
+    ssl_context.load_verify_locations(cafile=certifi.where())
```
zubenkoivan, line 60: Had to change this because locally I had SSL verification errors

asvetlov: ok, if it works

PR #3199, opened 2025-03-06
Platform Apps list instances

paul-gsd:


apolo-cli/src/apolo_cli/apps.py
```diff
@@ -0,0 +1,70 @@
...
+@argument("org", type=ORG, required=False)
+@argument("project", type=PROJECT, required=False)
+async def list(
+    root: Root,
+    cluster: Optional[str],
+    org: Optional[str],
+    project: Optional[str],
+) -> None:
+    """
+    List all app instances.
```
dalazx, line 32: it should be implied that `apps` are instances, in contrast to `app-templates`.

dalazx: so, I guess we should not use the term `instance` here.

apolo-sdk/src/apolo_sdk/_apps.py
```diff
@@ -0,0 +1,72 @@
...
+
+from ._config import Config
+from ._core import _Core
+from ._rewrite import rewrite_module
+from ._utils import NoPublicConstructor, asyncgeneratorcontextmanager
+
+
+@rewrite_module
+@dataclass(frozen=True)
+class AppInstance:
```
dalazx, line 12: same.

apolo-cli/src/apolo_cli/apps.py
paul-gsd: so that the command looks like
`apolo app-instances list` ?

dalazx: no why. `apolo app list`.

dalazx: let's *not* use the term `instance`.

dalazx: rather `app-template` and `app`.

paul-gsd: Ah, okay

paul-gsd:
Screenshot 2025-03-06 at 14 31 17
Example output

apolo-cli/src/apolo_cli/apps.py
```diff
@@ -0,0 +1,61 @@
...
+    table.add_column("Version")
+    table.add_column("State")
+
+    count = 0
+    async with client.apps.list(
+        cluster_name=cluster, org_name=org, project_name=project
+    ) as it:
+        async for app in it:
+            count += 1
+            table.add_row(
```
asvetlov, line 46: Please use a formatter instead of direct making the table, as it is done here: https://github.com/neuro-inc/apolo-cli/blob/master/apolo-cli/src/apolo_cli/secrets.py#L80-L90

Later formatter could be modified to hide/show additional fields, etc. For example, jobs formatter has rich and complex logic: https://github.com/neuro-inc/apolo-cli/blob/master/apolo-cli/src/apolo_cli/formatters/jobs.py#L75

Formatter tests should be done by `rich_cmp(console)` usage, as here: https://github.com/neuro-inc/apolo-cli/blob/master/apolo-cli/tests/unit/formatters/test_jobs_formatters.py#L256

To regen ascii files use `pytest ... --rich-gen` option, then check diffs for generated files

apolo-sdk/src/apolo_sdk/_apps.py
```diff
@@ -0,0 +1,72 @@
...
+    ) -> AsyncIterator[App]:
+        cluster_name = cluster_name or self._config.cluster_name
+        if org_name is None:
+            org_name = self._config.org_name
+            if org_name is None:
+                raise ValueError("Organization name is required")
+        if project_name is None:
+            project_name = self._config.project_name
+            if project_name is None:
+                raise ValueError("Project name is required")
```
asvetlov, line 44: If I directly pass, say, `org_name`, I cannot use default `project_name`, isn't it?
Please carefully process all possible argument combinations where defaults are applicable and where they are not.

asvetlov (changes requested):
Looks good in general, but please improve a couple things.

PR #3222, opened 2025-04-04
Update release instructions for changelog entry creation and announcements

paul-gsd:


RELEASE.md
```diff
@@ -16,6 +17,5 @@
 * Open PyPI (https://pypi.org/project/apolo-sdk/ and https://pypi.org/project/apolo-cli/),
   make sure that a new release is published and all needed files are awailable for downloading
   (https://pypi.org/project/apolo-sdk/#files and https://pypi.org/project/apolo-cli/#files).
-* Merge created pull request for `platform-web` project to publish updated documentation on the web.
 * Increment version to next alpha, e.g. `__version__ = 20.6.23a0`. Commit this change to master and push on github.
```
paul-gsd, line 20: @asvetlov do we need to do it, is it actual?

asvetlov: It would be good but not necessary

asvetlov: ok

asvetlov (approved):
Please merge when ready

PR #3264, opened 2025-05-30
extend job description with the namespace

parikls:


apolo-sdk/src/apolo_sdk/_jobs.py
```diff
@@ -252,6 +252,7 @@ class JobDescription:
     privileged: bool = False
     priority: JobPriority = JobPriority.NORMAL
     energy_schedule_name: Optional[str] = None
+    namespace: Optional[str] = None
```
zubenkoivan, line 255: may let's make it required with `platform-jobs` as default value?

zubenkoivan (approved):

PR #3320, opened 2025-08-15
Add force option to app uninstall command and update related documentation

paul-gsd:


apolo-sdk/src/apolo_sdk/_apps.py
```diff
@@ -155,6 +155,7 @@ async def uninstall(
         cluster_name: Optional[str] = None,
         org_name: Optional[str] = None,
         project_name: Optional[str] = None,
+        force: bool = False,
```
asvetlov, line 158:
```suggestion
        *,
        force: bool = False,
```

PR #3506, opened 2026-08-07
IT-88: fix acl ls help wording and SCREAMING_SNAKE_CASE rendering in generated docs

yarka-guru:
Two documentation defects found while writing the Console-side service accounts docs for IT-88 ([apolo-documentation#9](https://github.com/neuro-inc/apolo-documentation/pull/9)). Both surface on published pages under `docs.apolo.us/index/apolo-cli/`.

## 1. `apolo acl ls` help had its two sentences swapped

The docstring claimed the default lists resources shared **BY** you and `--shared` lists resources shared **WITH** you. The implementation is the reverse:

- default → `users.get_acl(username)`, two columns (URI, action) — what the user can reach;
- `--shared` → `users.get_shares(username)`, three columns (URI, action, **user**) — what the user has shared, and with whom.

The `--shared` flag's own help string ("Output the resources shared by the user") already agreed with the implementation, so only the docstring prose was wrong. Reworded to match, and to mention the third column, since that is the reason to reach for the flag.

Verified against a live cluster: `apolo acl ls --shared` prints the grantee column, `apolo acl ls` does not.

## 2. `SCREAMING_SNAKE_CASE` identifiers were split across several code spans

`site-help-generator.py` wrapped runs matching `[A-Z0-9\-]{3,60}` in backticks. Underscore is not in that class, so every underscore-separated run got its own pair:

```
- a full passed-config token, which can be used as ``APOLO`_`PASSED`_`CONFIG``
Get service account `SERVICE`_`ACCOUNT`.
```

The identifier is now matched as a whole. This fixes six pages, not just `service-account` — `APP_ID`, `REVISION_NUMBER`, `BUCKET_CREDENTIAL`, `CLUSTER_NAME`, `ORG_NAME`, `PROJECT_NAME` and `DISK_ID` were all affected.

The regex also now consumes already-quoted spans and passes them through, so a docstring that quotes an identifier itself (as `service-account create` does) no longer comes out double-quoted as ` ``APOLO_PASSED_CONFIG`` `. The transform is idempotent, and every non-identifier case it handled before is byte-for-byte unchanged — including mixed-case text like `FOO_bar`, which still yields `` `FOO`_bar ``.

## Regenerated output

`apolo-cli/docs/*.md` and `CLI.md` are regenerated with `make docs` (`markdown-toc==1.2.6` from `requirements/mypy.txt`). The only `CLI.md` change is the `acl ls` paragraph, which confirms the local toolchain reproduces the committed output rather than reformatting it.

## Testing

`pytest apolo-cli/tests/unit -k "share or acl or service_account or help or docs"` gives **17 failed, 24 passed** both with and without this branch — the failures are pre-existing in my local py3.14 / click 8.4 environment and are unaffected by these changes. Leaving CI to be the real check.

## Not fixed here

`click.wrap_text` breaks at hyphens, so `passed-config` in the `service-account create` docstring still renders as "a full passed- config token" after the line rejoins. That is a separate wrapping issue affecting any hyphenated word landing on a wrap boundary; happy to take it in a follow-up if you want it.

### DIFF
--- a/CHANGELOG.D/3506.doc
+++ b/CHANGELOG.D/3506.doc
@@ -1,0 +1,1 @@
+Stopped the docs generator from splitting `SCREAMING_SNAKE_CASE` identifiers across several code spans, so metavars such as `APOLO_PASSED_CONFIG`, `SERVICE_ACCOUNT` and `APP_ID` now render as one.
--- a/CHANGELOG.D/3506.bugfix
+++ b/CHANGELOG.D/3506.bugfix
@@ -1,0 +1,1 @@
+Fixed `apolo acl ls` help text, which described the default output and the `--shared` output the wrong way round.
--- a/CLI.md
+++ b/CLI.md
@@ -371,7 +371,7 @@ Name | Description|
 
 ### apolo acl ls
 
-List shared resources.<br/><br/>The command displays a list of resources shared BY current user \(default).<br/><br/>To display a list of resources shared WITH current user apply --shared option.<br/><br/>Use `\-u/--username` to fetch roles of a specified user or role.<br/>
+List shared resources.<br/><br/>The command displays a list of resources available TO current user \(default).<br/><br/>To display a list of resources shared BY current user, along with the users<br/>and roles they are shared with, apply --shared option.<br/><br/>Use `\-u/--username` to fetch roles of a specified user or role.<br/>
 
 **Usage:**
 
--- a/apolo-cli/docs/acl.md
+++ b/apolo-cli/docs/acl.md
@@ -125,14 +125,15 @@ apolo acl ls [OPTIONS] [URI]
 
 List shared resources.
 
-The command displays a list of resources shared BY
+The command displays a list of resources available TO
 current user (default).
 
-To display a list of resources shared `WITH` current
-user apply --shared option.
+To display a list of resources shared BY current
+user, along with the users and
+roles they are shared with, apply --shared
+option.
 
-Use `-u/--username` to fetch roles of a specified
-user or role.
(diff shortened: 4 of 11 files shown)


CHANGELOG.D/3506.doc
```diff
@@ -0,0 +1 @@
+Stopped the docs generator from splitting `SCREAMING_SNAKE_CASE` identifiers across several code spans, so metavars such as `APOLO_PASSED_CONFIG`, `SERVICE_ACCOUNT` and `APP_ID` now render as one.
```
YevheniiSemendiak, line 1: I'm not sure we have `*.doc` support in towncrier template. You can rename this file to `3506.misc`.

YevheniiSemendiak (approved):

CHANGELOG.D/3506.doc
yarka-guru: Перевірив — `doc` є серед дефолтних типів towncrier, окремої конфігурації не потребує. `towncrier build --draft` на цій гілці дає:

```
## Bugfixes

- Fixed `apolo acl ls` help text, which described the default output and the `--shared` output the wrong way round. ([#3506](...))

## Improved Documentation

- Stopped the docs generator from splitting `SCREAMING_SNAKE_CASE` identifiers across several code spans... ([#3506](...))
```

Тобто фрагмент підхоплюється і потрапляє в секцію "Improved Documentation", що тут точніше за "Miscellaneous". Лишив як є — але якщо хочеш тримати репозиторій лише на `feature`/`bugfix`/`misc`, скажи, перейменую.

PR #3522, opened 2026-08-21
APS-412: Let the CLI apply a config written for another version

yarka-guru:
Closes [APS-412](https://apolocloud.atlassian.net/browse/APS-412), the CLI half of [APS-405](https://apolocloud.atlassian.net/browse/APS-405).

## Problem

`apolo app configure` refused any file whose `template_version` differed from the installed app's:

```
$ apolo app configure <id> -f config.yaml
ERROR: Cannot update app: template name or version mismatch
```

The version in the file was never used for anything — `configure` sends only `display_name` and `input`, and the app stays on the version it was installed with. The console dropped the same comparison in APS-405 (#605), where measuring the catalogue showed 58% of consecutive version pairs ship a byte-identical schema and 35 of 36 templates publish moving refs like `master`, so the string tells you nothing.

## Change

The version comparison is gone. The template **name** is still checked, and says what it found:

```
ERROR: Cannot update app: it was installed from 'aws-s3' and the config is for 'shell'
```

## The part that is not just deleting a check

The app API answers **500 Internal Server Error** — not a validation error — when the input carries a field the template does not define:

```
PUT .../instances/<id>   valid input                          -> 200
PUT .../instances/<id>   same input + one undefined field     -> 500   (twice, deterministic)
POST .../instances       same input + one undefined field     -> 500
```

That is exactly what a config written for an older version usually carries: a field the newer version renamed or dropped. In the console's replay of the catalogue, about one cross-version config in four carries such a field, and `apolo app install -f` can hit this today — there is no version gate on install.

So dropping the gate alone would have swapped one bad experience for a worse one. `configure` now fetches the schema of the version the app runs and names those fields itself:

```
ERROR: Cannot update app: 26.0.0 has no s3.dockerconfigjson, gone_in_this_version.
Remove them from the config or install a version that still has them.
```

The check only reports fields the schema has no place for — everything else stays the server's business, and a template that cannot be fetched skips the check entirely rather than blocking the call.

I will raise the 500 separately against the app API; this is a client-side guard, not the fix.

## Verified on dev

Against a throwaway `aws-s3` instance installed on `26.0.0` (uninstalled afterwards):

| | before | after |
|---|---|---|
| config for `26.0.1`, app on `26.0.0` (identical schemas) | `template name or version mismatch` | **applies** |
| config for another app | same generic message | names both templates |
| config carrying `s3.dockerconfigjson` | would have been a 500 | names the field |

Read back with `apolo app get-input` after the cross-version configure:

```yaml
s3:
  bucket: aws412-reconfigured   # from the 26.0.1 file
  port: 9090
```

## Tests

Three cases against the fake server — cross-version configure sends `{display_name, input}` and nothing else, a config for another app is refused, a config with an undefined field is refused naming it — plus a unit test for the schema walk covering `$ref`/`$defs`, `anyOf` variants, nested paths, and schemas that declare no properties.

`pytest apolo-sdk` 767 passed, `mypy apolo-sdk` clean, black/isort/flake8 clean.

[APS-412]: https://apolocloud.atlassian.net/browse/APS-412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[APS-405]: https://apolocloud.atlassian.net/browse/APS-405?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.D/3522.feature
+++ b/CHANGELOG.D/3522.feature
@@ -1,0 +1,1 @@
+Applying a config written for another version of the same app with `apolo app configure` no longer fails on a version mismatch. Fields the installed version does not have are named instead.
--- a/apolo-sdk/src/apolo_sdk/_apps.py
+++ b/apolo-sdk/src/apolo_sdk/_apps.py
@@ -10,6 +10,7 @@ from yarl import URL
 
 from ._config import Config
 from ._core import _Core
+from ._errors import ResourceNotFound
 from ._rewrite import rewrite_module
 from ._utils import NoPublicConstructor, asyncgeneratorcontextmanager
 
@@ -97,6 +98,68 @@ class AppConfigurationRevision:
     comment: str | None
     created_at: datetime
     end_at: datetime | None
+
+
+def _resolve(schema: dict[str, Any], root: dict[str, Any]) -> list[dict[str, Any]]:
+    ref = schema.get("$ref")
+    if isinstance(ref, str) and ref.startswith("#/$defs/"):
+        target = root.get("$defs", {}).get(ref.split("/")[-1])
+        return _resolve(target, root) if isinstance(target, dict) else []
+
+    variants = schema.get("anyOf") or schema.get("oneOf")
+    if isinstance(variants, list):
+        return [
+            resolved
+            for variant in variants
+            if isinstance(variant, dict)
+            for resolved in _resolve(variant, root)
+        ]
+
+    return [schema]
+
+
+def undefined_input_fields(
+    schema: dict[str, Any] | None,
+    value: Any,
+    root: dict[str, Any] | None = None,
+    prefix: str = "",
+    depth: int = 0,
+) -> list[str]:
+    """Paths in ``value`` that ``schema`` does not define.
+
+    The app API answers 500 rather than a validation error when it is given a
+    field the template does not have, which is what a config written for
+    another version of the same app usually carries.
+    """
+    if not isinstance(schema, dict) or not isinstance(value, dict) or depth > 10:
+        return []
+
+    root = root if root is not None else schema
+    variants = _resolve(schema, root)
+
+    known: dict[str, dict[str, Any]] = {}
+    for variant in variants:
+        properties = variant.get("properties")
+        if isinstance(properties, dict):
+            for name, definition in properties.items():
+                if isinstance(definition, dict):
+                    known.setdefault(name, definition)
+
+    if not known:
+        return []
+
+    undefined = []
+    for name, nested in value.items():
+        path = f"{prefix}.{name}" if prefix else name
+        definition = known.get(name)
+        if definition is None:
+            undefined.append(path)
+        else:
+            undefined.extend(
+                undefined_input_fields(definition, nested, root, path, depth + 1)
+            )
+
+    return undefined
 
 
 @rewrite_module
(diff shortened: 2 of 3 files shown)


yarka-guru:
## Second opinion from Codex, and what it changed (`8c4d75dc`)

Ran this diff past Codex as a second reviewer, then checked each finding against the real catalogue schemas before believing any of it. Three of five were real, and the catalogue says two of those are the common case rather than an edge:

| finding | occurrences in the catalogue | effect |
|---|---|---|
| array elements were never walked | **188** arrays of objects | a stale key inside one still reached the API and hit the 500 this exists to prevent |
| a property several union variants define was read against the first one only | **305** such properties | a key valid under a later variant was reported undefined, refusing a config that would have applied |
| schemas accepting extra keys treated as closed | **31** use `additionalProperties` | legal keys reported too |

All three are fixed, each with a test. Behaviour now:

```
real config vs its own version   []
renamed field vs v26.8.0         ['image.dockerconfigjson']
stale key inside an array        ['networking.ports.0.old_flag']     <- was missed
key from the second variant      []                                  <- was a false positive
key under additionalProperties   []                                  <- was a false positive
```

Codex also pointed out that only a 404 from the template read was tolerated, so a 403, a timeout or a 500 there would have refused a call the server would have accepted. Any failure to read the schema now skips the check — it is a courtesy and must never stand between the user and a valid call.

Not changed, deliberately: `allOf`, `patternProperties` and refs outside `#/$defs/` are unhandled, and the catalogue has **zero** of each. Their failure mode is a missed field, which is where we already were.

771 SDK tests, mypy/flake8/isort/black clean.

yarka-guru:
## Reviewing my own diff (`76bdfb0b`)

Nine of fifteen findings were real, and the first one was a regression I would have shipped.

**A value taken from another app was rejected.** `apolo app-template get-config` writes a header telling users to replace any parameter with

```yaml
my_param:
  type: "app-instance-ref"
  instance_id: "<app-instance-id>"
  path: "<path-from-get-values-response>"
```

The check read those three keys as fields the schema does not define and refused the config — breaking a documented feature that worked before this branch. Such a value is left alone now.

**A schema pointing at itself crashed the command.** `_resolve` followed `$ref` with no memory of where it had been; a recursive alias gave a RecursionError traceback. Worse, the call sat *outside* the guard whose comment promises it will never stand between the user and a valid configure. Both fixed — refs carry the set already followed, and the check runs inside the guard.

**Arrays were read against the first variant only.** For `list[X] | None` with the null variant first — the ordinary shape — `variants[:1]` picked `{type: null}`, whose `items` is None, so the array was never checked and the 500 came through anyway. That was my own half-fix from the previous round: the dict path intersects across variants, the list path did not. It does now.

**`allOf` wrappers and `#/definitions/` pointers were skipped.** I had dismissed both after counting zero in the dev catalogue, but pydantic emits `allOf` for any field carrying a description, and this repo's own `template_schema_utils.py` has always resolved both — so the two walkers disagreed about whether a config gets checked at all. Handled.

Also in: the check is skipped when the config names the version the app already runs, so the ordinary path keeps its single request and none of the new logic; a config that names no template is accepted (nothing downstream needs it) instead of being told it is for `None`; the swallowed template error is logged at debug; array paths sort by index rather than lexically (`ports.2` before `ports.10`); the message names the template and caps a long list; and the helper is private with the recursion parameters no longer part of a public signature.

Checked afterwards: **317 catalogue schemas report nothing against their own shape**, and each new test fails if the behaviour it covers is removed.

776 SDK tests, mypy/flake8/isort/black clean.

Left alone deliberately: `patternProperties` (zero in the catalogue, and its failure mode is a missed field, not a false refusal) and unifying the two schema walkers — the CLI's resolver is the better one and the SDK cannot import from the CLI, so lifting it is its own change rather than a rider on this.

sasha-apolo (approved):

